PR #1, opened 2020-07-08
Minor corrections in the openlane configs

ax3ghazy:
- typos corrected
- capacitive load for the hs scl corrected


RTimothyEdwards:
Pulled.  Thanks for the updates!

PR #6, opened 2020-07-17
common: Fix mixed tab/space in foundry_install.py

smunaut:
Python doesn't like mixing tab and space indentation.
Replaced every tab with 4 spaces.

Signed-off-by: Sylvain Munaut <tnt@246tNt.com>

### DIFF
--- a/common/foundry_install.py
+++ b/common/foundry_install.py
@@ -7,11 +7,11 @@
 # should be a staging area, not a place where files are kept permanently.
 #
 # Options:
-#    -ef_format		Use efabless naming (libs.ref/techLEF),
-#			otherwise use generic naming (libs.tech/lef)
-#    -clean		Clear out and remove target directory before starting
-#    -source <path>	Path to source data top level directory
-#    -target <path>	Path to target (staging) top level directory
+#    -ef_format         Use efabless naming (libs.ref/techLEF),
+#                       otherwise use generic naming (libs.tech/lef)
+#    -clean             Clear out and remove target directory before starting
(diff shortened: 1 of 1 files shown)


RTimothyEdwards:
How does this fail for you?  I have been using this script on many different systems without any problems.  I only have tabs on continuation lines and python never complains about it.

smunaut:
```
set -f ; ../common/foundry_install.py -std_format -source ~/gits/osu_130_pdk -target `pwd`/sky130A \
	-techlef char/techfiles/scs8.lef rename sky130_osu_sc.tlef \
	-spice lib/spice/*.spice compile-only \
	-lef outputs/s8_osu130.lef rename sky130_osu_sc.lef \
	-lib outputs/*.lib \
	-gds lib/gds/*.gds compile-only \
	-verilog outputs/VERILOG/*.v \
	-library digital sky130_osu_sc
  File "../common/foundry_install.py", line 617
    print('gds library', file=ofile)
                                   ^
TabError: inconsistent use of tabs and spaces in indentation
```

smunaut:
Looks like a commit from yesterday introduced that.

RTimothyEdwards:
Yep, I really do know better than that. . .  Should not make and edit and commit without testing!  I am doing way too much multitasking.  Anyway, thanks for the fix.  I just fixed and pushed and then realized yours was a pull request and not an issue, so sorry for bypassing your pull request. . .

RTimothyEdwards:
FYI, I just pushed to the opencircuitdesign.com repo and the github repo gets mirrored overnight.

PR #8, opened 2020-07-27
initial hvl support for openlane

agorararmard:
These are the necessary configurations for hvl to get it to run on openlane.
The sky130/Makefile was also changed to include those files.


RTimothyEdwards:
@agorararmard :  Why did you close this?

agorararmard:
@RTimothyEdwards We have many other updates that we're preparing, so we thought it's better for all the updates to come on several pull requests but from the same source and in the order in which we applied them, which will be @ax3ghazy

PR #9, opened 2020-07-27
OpenLane support for other sky130 variants

ax3ghazy:
- Added support for ls, ms, hdll, and hvl
- power nets are named after the VDD and GND pin names; makes LVS report super
  clean
- Added support for both EF_STYLEs in OpenLane configuration files


RTimothyEdwards:
@ax3ghazy :  This presumably all works, but I question why in your files you sometimes use env(PDK_VARIANT) and sometimes have the PDK variant name written out.  In other places you somtimes have env(PDK) and in other places TECHNAME.

For example of the former, if you have `$::env(PDK_VARIANT)/lib/sky130_fd_sc_hd__ff_n40C_1v95.lib` then the environment variable is essentially meaningless;  if you change PDK_VARIANT then everything breaks, because there will be no file "sky130_fd_sc_hd__..." in any location except PDK_VARIANT=sky130_fd_sc_hd.

For an example of the latter, if you have `set ::env(MAGIC_MAGICRC) "$::env(PDK_ROOT)/$::env(PDK)/libs.tech/magic/20200508/TECHNAME.magicrc"` then the environment variable is also essentially meaningless, because if you change PDK to anything other than TECHNAME then there will be no such TECHNAME.magicrc in that path.

I also disagree with the use of the name "PDK_VARIANT" to refer to an IP library, which is not a PDK variant.

ax3ghazy:
In the first case, strings like `$::env(PDK_VARIANT)/lib/sky130_fd_sc_hd__ff_n40C_1v95.lib` only occur in paths like `sky130A/libs.tech/openlane/**sky130_fd_sc_hd**/config.tcl`, in which the standard cell library is already determined by the path, so it's guaranteed if this file is being read, PDK_VARIANT must be correct. I don't see an advantage in these cases to replace $::env(PDK_VARIANT) by the actual variant name for each of the variants. Let me know if you think otherwise.

The second case is similar since they all exist in the configuration that is common among all standard cell libraries; hence, it exists under `**sky130A**/libs.tech/openlane/config.tcl`, where PDK must be set to sky130A (TECHNAME) if this file is being accessed by openlane. 

However, since `open_pdks` is what actually generates the whole installation directory; I will probably go ahead and stick to the "lingo" of `open_pdks` and exclusively use TECHNAME and LIBRARY like in qflow. Do you find that cleaner?

RTimothyEdwards:
@ax3ghazy :  Actually I was thinking the other way around:  You use "env(PDK_VARIANT)" or "env(PDK)" in *both* places in the path, so that if you change "env(PDK_VARIANT)" everything switches automatically to a different standard cell set;  likewise, should there be more variants than just "sky130A", then changing "env(PDK)" would switch automatically to the other PDK variant (e.g., "sky130B";  this makes more sense if you think about it with the efabless XFAB varieties like switching from EFXH018A to EFXH018D).

RTimothyEdwards:
Sorry for sitting on this for a couple of days.  I have pulled and merged.

ax3ghazy:
Thanks!

PR #10, opened 2020-07-27
Minor typos fixed in the Makefile

ax3ghazy:
- rename needs '='
- *.*.v -> *.*.lef (to ignore .magic.lef files)


RTimothyEdwards:
All looks good;  merging.

PR #11, opened 2020-07-27
Setting the default LINK_TARGETS to none

ax3ghazy:
- The installation can run in docker containers by default of course at the cost
  of disk space.


RTimothyEdwards:
Agreed, LINK_TARGETS should be set to "none" for the default behavior.  This has been merged.

PR #12, opened 2020-07-27
Minor fixes to the magicrc and netgen setup files

ax3ghazy:
- netgen regex fixed to ignore fill cells and decaps
- typos fixed

### DIFF
--- a/sky130/magic/sky130.magicrc
+++ b/sky130/magic/sky130.magicrc
@@ -16,7 +16,7 @@ if {[lindex $scalefac 1] < 2} {
 drc euclidean on
 
 # Allow override of PDK path from environment variable PDKPATH
-if {[catch {set PDKPATH $env(PDKPATH)}]} {
+if {[catch {set PDKPATH $::env(PDKPATH)}]} {
     set PDKPATH "STAGING_PATH/TECHNAME"
 }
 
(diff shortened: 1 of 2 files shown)


RTimothyEdwards:
My belief is that the fill and decap cells should not be ignored, and that critically, the tap cells must not be ignored, but that is all dependent on the problems caused by the SkyWater cells having the VNB and VPB pins on the wrong layer.  Once we have properly fixed the standard cell views, this probably needs to be revisited.  Overall it is better never to ignore a cell, otherwise you can miss errors that are caused by the ignored cell itself, such as an improper overlap or something.

However, for now, I am merging this.

ax3ghazy:
Thanks; agreed. I thought, however, that the tap connection is _only_ done below the local interconnect and would make the tap cell basically like a fill cell or a decap if extraction was done using the maglef views. Is that correct?

RTimothyEdwards:
The change I made to the LEF and magic techfile, adding MASTERSLICE entries for "pwell" and "nwell", means that the .maglef view would have "VNB" connected to a square of "pwell" (which then is extracted as substrate) while "VPB" will be connected to nwell, which being a masterslice layer, will appear in the .maglef view as an nwell layer (which also solves the problem of making sure magic checks nwell spacing between standard cells and everything else).  So the connectivity of the substrate, well, and power supplies will all be correctly and exactly represented in the .maglef views and will extract correctly, and the tap cells will correctly merge the bias connections to the power rails.

PR #15, opened 2020-07-27
Linking from None fix

ax3ghazy:
- Shouldn't do the symbolic linking if link_from is None


RTimothyEdwards:
This got fixed in the interim.  Your version is probably more "pythonic".  I am leaving it with my fix unless you have an issue with it.

ax3ghazy:
Oh, I just haven't noticed it was already fixed. No issue with the current fix; closing this.

PR #16, opened 2020-07-27
Commented out missing IO and OSU libraries

ax3ghazy:
Hi @RTimothyEdwards,
You might probably want to handle this in a different way? Would something like this: https://github.com/google/skywater-pdk/blob/master/Makefile#L68 make sense here?

### DIFF
--- a/sky130/Makefile
+++ b/sky130/Makefile
@@ -348,30 +348,30 @@ vendor-a:
 		-library digital sky130_fd_sc_ms \
 		-library digital sky130_fd_sc_lp |& tee -a ${SKY130A}_install.log
         # Install OSU digital standard cells.
-	${STAGE} -source ${OSU_PATH} -target ${STAGING_PATH}/${SKY130A} \
-		-techlef char/techfiles/scs8.lef rename sky130_osu_sc.tlef \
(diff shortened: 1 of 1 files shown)


RTimothyEdwards:
I was "handling this isn a different way" by ignoring it and letting it generate errors for those libraries until such time as they would be available.  Which I thought would be before now, for both of them.  Since there have already been comments on the slack channel about open_pdks throwing these errors, I'll go ahead and apply the comments.  Merged and closed.

ax3ghazy:
I was rather asking if you think my "comment them out" solution should be handled in a cleaner way, but i guess that would be a wasted effort once they're released. Maybe I can try adding support for specific targets some when later (e.g., `make sky130_fd_sc_hd`) (that's why I linked that line on skywater-pdk).

Thanks.

RTimothyEdwards:
@ax3ghazy :  Per another comment I made about a "configure" script, I think a cleaner way might be to use "configure" with a set of defaults so that you can do "configure --without-sky130_fd_sc_hd" or something like that.  Honestly, there's a point at which it's better to drive the whole thing with a GUI.

PR #17, opened 2020-07-27
PDK_ROOT needs to be set

ax3ghazy:
Hello,

This might be too openlane-specific, but this forces/allows the user to set the installation path and the skywater-pdk installation path without editing the Makefile.

Other alternatives include checking if LOCAL_PATH and SKYWATER_PATH are already set externally and use the values provided by the environment.

Please let me know what you think.

Thanks,
Ahmed

-----------------------------
- PDKs (e.g., skywater-pdk) are assumed to be under $PDK_ROOT
- Default LOCAL_PATH is set to $PDK_ROOT


RTimothyEdwards:
There should not be a requirement that the location of the source skywater-pdk directory and the local install location must be rooted at the same point.

There is a use in having the locations set by something like a configure script, so that such things would be specified by "configure --help" and passed by "configure --prefix=", which is the normal way to do it instead of telling people to go edit the Makefile.

ax3ghazy:
Agreed about both points; fair enough. I will re-do this with `autotools`; need at least a `--prefix=`/`--local-path=` and something like `--pdk-source[-techname]=`.

RTimothyEdwards:
That's a rather large undertaking. . . particularly as once you go down that path, it makes sense to do a lot of other things like auto-detect the presence of magic or openlane or skywater-pdk and set all the path pointers to the sources automatically. . .  But it does solve another problem, which is that changes can then be made to the master configure.ac script and the Makefile can be put in .gitignore, so locally you don't need to mess with having the Makefile get overwritten every time you do a git pull.

ax3ghazy:
I have addressed this alternative in https://github.com/RTimothyEdwards/open_pdks/pull/19.

RTimothyEdwards:
I am a about to apply the pull request #19.  Should this issue just be closed, then?

ax3ghazy:
Yup, I am closing this. Thanks.

PR #18, opened 2020-07-27
Renamed tech name in klayout .lyt file to sky130A

ax3ghazy:


RTimothyEdwards:
The underlying problem is that the file, like most files processed by the Makefile, is supposed to not have the PDK name in it at all.  The PDK name should be replaced by the bareword TECHNAME:

`${CPP} ${SKY130A_DEFS} klayout/${TECH}.lyp > ${KLAYOUT_STAGING_A}/${SKY130A}.lyp`

where `SKY130A_DEFS = -DTECHNAME=sky130A -DREVISION=${REVISION}` (plus other stuff).

ax3ghazy:
Ah, got it. Updated the commit with TECHNAME instead.

RTimothyEdwards:
Looks good;  merged and closed.

PR #21, opened 2020-08-08
Continuation operator needed in sky130/Makefile.in

ax3ghazy:
Every line is run in a new shell by make, so need a continuation operator there.


RTimothyEdwards:
Ah, thanks for catching and testing that.  Pulled and merged.

PR #23, opened 2020-08-10
Small whitespace fixes

mithro:
* Remove all trailing spaces.
 * Fix stray tab.


mithro:
This has your three changes in it because of the very weird way you are using git / github.

RTimothyEdwards:
To avoid problems with merging in the changes that have already been made but not propagated to github, I have made the whitespace fixes locally and they will go in with tonight's mirror update.

PR #30, opened 2020-09-29
openlane configs updates and cleanups

agorararmard:
- no_synth.cells cleanup
- first go at no_synth.cells reduction
- variables cleanup
- core_ring support


RTimothyEdwards:
Pulled and merged.

PR #32, opened 2020-10-07
Mod lefs

agorararmard:
- An attempt at modifying the LEF (.magic.lef) files (VNB and VPB) in the skywater-pdk before installing them.
- Also, adding the missing info from the hs and ms diodes.
- Modifying openlane configs to finally use hs and ms diodes.
Not the best implementation, but: an implementation.


RTimothyEdwards (changes requested):
Nothing foundry-specific should ever be placed in common/.  "common" means something common to all foundries.  Foundry-specific scripts go in, e.g., "sky130/custom/scripts/" or anywhere down the "sky130/custom/" path.

However, if there are any serious errors in the SkyWater repository, then we will fix those in the SkyWater repository.  Errors that seem to be long-running can be fixed by custom scripts in sky130/custom/.  However, in this case I am aiming for a quick turnaround fixing the PDK.  I have already corrected the scripts that generate the LEF files based on your observations, and I hope to have the standard cell libraries updated in a day or two.

common/fix_missing_lef.py
```diff
@@ -0,0 +1,81 @@
...
+args = parser.parse_args()
+skywater_path = args.skywater_path
+modify_hs_ms_diodes = args.modify_hs_ms_diodes
+modify_vnb_vpb = args.modify_vnb_vpb
+
+libraries = ['sky130_fd_sc_hd','sky130_fd_sc_hdll','sky130_fd_sc_hs','sky130_fd_sc_ms','sky130_fd_sc_ls']
+
+if modify_vnb_vpb:
+    for lib in libraries:
+        revert_modify_vpb_result_cmd = "sed -i 's/PIN VPB\\n    DIRECTION INOUT ;\\n    USE POWER ;/PIN VPB/g' {lef} ".format(
```
mithro, line 29: Every one of these fixes should be referencing a bug on the google/skywater-pdk GitHub repository.

agorararmard: @mithro: For the vnb and vpb pins, I communicated it with Tim Edwards, and as he said they will be updated in a day or two, so I felt there is no need to open an issue for this. However, I can open one just for the sake of traceability.

As for the ms and hs diodes, this has been a long-standing issue, for a couple of months now: https://github.com/google/skywater-pdk/issues/141

I'm sure they will eventually be fixed in the original PDK source, however, these are just an "until-then" solutions.

agorararmard:
@RTimothyEdwards: Since the vnb and vpb issue will be fixed that soon, I will remove its fix from the script: it will only modify the ms and hs diode info. 
I will definitely move the script to `sky130/custom/scripts` (sorry about that). 
Of course, I agree that these fixes should go into the Skywater repo, but as I said, this is just an "until-then" solution.

RTimothyEdwards:
@agorararmard :  I pulled and merged #30 .   However, I do recall having merge conflicts in that one, as well, which I thought I resolved.

common/fix_missing_lef.py
mithro: Couple of things;
 * If there is no github issue tracking something, then it doesn't exist.
 * Adding a simple comment like "This can be removed when GitHub issue XXXX is fixed." is perfectly fine.

agorararmard: Roger that.

agorararmard: I added the comment: https://github.com/RTimothyEdwards/open_pdks/pull/32/commits/fcedadef081398e7742a8ddf23d2db9d52965427
And opened a bug report to track the vnb and vpb issue: https://github.com/google/skywater-pdk/issues/172

agorararmard:
@RTimothyEdwards: I guess the latest 2 commits cover the changes that you requested, am I right?

RTimothyEdwards (approved):
Since I haven't seen any movement on this in the skywater-pdk repository, I will go ahead and merge this into open_pdks.  I would like to make sure it is removed when/if the issue is finally taken care of in skywater-pdk.

RTimothyEdwards:
Pull request has been merged.

PR #34, opened 2020-10-08
Follow up on 6f4d30: correctly install latch_map.v

ax3ghazy:
Previously forgot to update the Makefile to copy latch_map.v into the staging area. Do you think that using something like a find command to run all non-binary files through the preprocessor and copy the rest verbatim makes sense? If so, I can give it a try sometime.

### DIFF
--- a/sky130/Makefile.in
+++ b/sky130/Makefile.in
@@ -313,42 +313,54 @@ openlane-a: openlane/common_pdn.tcl openlane/config.tcl openlane/sky130_fd_sc_hd
 	rm -f ${OPENLANE_STAGING_A}/sky130_fd_sc_hd/tracks.info
 	rm -f ${OPENLANE_STAGING_A}/sky130_fd_sc_hd/no_synth.cells
 	rm -f ${OPENLANE_STAGING_A}/sky130_fd_sc_hd/sky130_fd_sc_hd__fakediode_2.gds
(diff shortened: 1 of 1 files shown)


RTimothyEdwards:
@ax3ghazy :  Check if you can use a "filter" script for that.  I'm not sure if that would work or not.  I like to run things through foundry_install.py whenever possible, and avoid shell scripts, mostly because even for as long as I've worked with them, I find shell scripts much more difficult to read and understand than a well-written python program.

RTimothyEdwards:
@ax3ghazy :  Is this pull request ready to merge?

ax3ghazy:
I am working on an alternative; I will close this for now since it in conflict with the base branch. I agree that things like the following, though concise, aren't very readable.

```
	mkdir -p ${OPENLANETOP_STAGING_A}
	mkdir -p ${OPENLANE_STAGING_A}
	for file in $$(find openlane -type f -printf '%P\n'); do\
		path_to_staging_file=${OPENLANE_STAGING_A}/$$file;\
		mkdir -p $$(dirname $$path_to_staging_file);\
		rm -f $$path_to_staging_file;\
		if file openlane/$$file | grep -q "ASCII"; then\
			${CPP} ${SKY130A_DEFS} openlane/$$file > $$path_to_staging_file;\
		else\
			cp openlane/$$file $$path_to_staging_file;\
		fi;\
	done
```

PR #35, opened 2020-10-14
Fix an indentation in cdl2spi.py

ax3ghazy:


RTimothyEdwards:
I already fixed this on my end yesterday, but I have not pushed that yet.  I will do so in a little bit, so I'm going to go ahead and close this pull request.  Thanks for going ahead and fixing it, though, since I left it in a broken state. . .

PR #38, opened 2020-10-21
fix the swapped lines in fix_missing_diodes_lef.py

agorararmard:
@ax3ghazy noticed that there are swapped lines in the script that would cause the build to fail if the user is not building the hs and ms libraries. So, here is the quick fix.
Sorry for that :-)


RTimothyEdwards:
Pull request has been merged.

PR #39, opened 2020-10-21
don't run mv unless the file exists

agorararmard:
The `mv` command that was used to rename the `.tlef` file was breaking the make, if the file didn't exist. So, I just added a check before running the `mv` to make sure the file exists. (the quick solution).


RTimothyEdwards:
Pull request has been merged.

agorararmard:
@RTimothyEdwards: The CI build is currently failing because this PR hasn't been merged yet, and so if you try to install the pdk without the osu_t18 being there the build will fail/break at the `mv` command.

PR #40, opened 2020-10-21
base commit for travisCI auto testing

agorararmard:
This is a working travisCI testing setup for open_pdks. This should check all pushes and pull requests to make sure that nothing breaks the build.

currently, the jobs are:
1. build hd library alone
2. hs
3. ms
4. hdll
5. ls
6. build all libraries together.

More jobs should be added when more libraries are added to the skywater-pdk public repo.

What it does:
1. build a magic docker container.
2. clone skywater-pdk and update the desired submodule
3. build open_pdks (invoke configure and make, from within the docker container)
4. Check if any fatal error happened during the build.
5. Check if the produced files have a sensible size (bigger than 250MB)

What you need to do to activate this from your side (@RTimothyEdwards):
1. go to https://travis-ci.com signup with your github account and link it.
That's it, enjoy the dashboard and the green ticks.

I'll be maintaining it every once in a while, if anything needs to be changed or added to the setup.

One downside of it is that you'll have to email support@travis-ci.com and tell them:
```
Dears,
   One job in my test-suite is long by nature because it builds multiple libraries using the tool in this repo: https://github.com/RTimothyEdwards/open_pdks

So, Could you increase the timeout for a job? Because 50 minutes is too short for us, and we cannot reduce the runtime in any way. Also, if you could increase the maximum job log to 5 or 6 MB.

My account name is: RTimothyEdwards
```

This is because the build process of job 6 (all the libraries together) takes about an hour, and produces a log that is bigger than 4MB (the maximum allowed).
For now, I set job 6 as an (allow_failures) job because it will eventually timeout, but at the same time, it would be useful to see if it broke/failed before it times out.

This tests the version on github only, and is activated on PRs and pushes to any of the branches.

Let me know if you want me to change/enhance/update anything!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,9 @@ scripts/*
 **/Makefile
 **/Makefile.am
 
+# exceptions for travisCI
+!.travisCI/Makefile
+
 # staged PDKs
 sky130/sky130A
 sky130/sky130A/*
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,46 @@
+# Copyright 2020 Efabless Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+language: minimal
+
+
+services:
+    - docker
+
+os:
+  - linux
+
+jobs:
+  include:
+    - name: "hd library"
+      env: STD_CELL_LIBRARY=sky130_fd_sc_hd
+    - name: "hs library"
+      env: STD_CELL_LIBRARY=sky130_fd_sc_hs
+    - name: "ms library"
+      env: STD_CELL_LIBRARY=sky130_fd_sc_ms
+    - name: "ls library"
+      env: STD_CELL_LIBRARY=sky130_fd_sc_ls
+    - name: "hdll library"
+      env: STD_CELL_LIBRARY=sky130_fd_sc_hdll
+    - name: "all libraries"
+      env: STD_CELL_LIBRARY=all
+  allow_failures:
+    - name: "all libraries"
+      env: STD_CELL_LIBRARY=all
+
+install:
+  - bash .travisCI/travisBuild.sh
+
+script:
+  - bash .travisCI/travisTest.sh
--- a/.travisCI/Dockerfile
+++ b/.travisCI/Dockerfile
@@ -1,0 +1,61 @@
+# Copyright 2020 Efabless Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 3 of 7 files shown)


mithro:
It is likely travis won't increase your limits, but you might be able to purchase the commercial travis plan which has longer timeouts and faster machines.

You might also want to consider GitHub actions which also have a longer time outs.

agorararmard:
@mithro: There have been changes (luckily for us) in travisCI. The commercial tool is now free for all public repos on Github (unlimited builds). We have this setup for openlane and build 490 (with a timeout of 3 hours for each job) just completed a few hours ago. The restriction would be on the number of parallel jobs (maximum 5), and since the first 5 jobs complete in 10-20 minutes, and only job 6 takes the hour, that wouldn't be a problem (probably the smart thing to do is to run the longest job first unlike what I did in this configuration file tho :-) ).

Anyhow, sending that email worked for OpenLANE, and it should work for open_pdks as well without requiring any subscriptions, as long as the repo remains public/open source.

RTimothyEdwards:
@agorararmard :  What is it that takes so long for the run?  A complete installation of libraries on my laptop takes about 10 minutes, and that includes the HVL, LP, PR, and IO libraries as well.

agorararmard:
@RTimothyEdwards: The free plan kinda has another caveat that we could live with, which is that each job (virtual machine) has 2 threads only, and so including the runtime of building the libraries in the skywater-pdk itself `make timing -j(nproc)` (which takes most of the time), the total runtime extends to about an hour.

RTimothyEdwards:
@agorararmard :  Make sure that you are only using the "latest" directories of each of the libraries.  open_pdks only builds on "latest", so you wouldn't want to do "make timing" on everything, or pull all submodule versions.

RTimothyEdwards:
@agorararmard :  I see from the travis-ci Makefile that you are, in fact, getting only the "latest" from the submodules.  Would there be any way to get a pre-built version of the skywater-pdks repository for this?

agorararmard:
@RTimothyEdwards: Yeah, that is the case. We're only building `latest` of each submodule, but the threads thing is what slows the process down for the (all libraries) job, the other jobs finish relatively in a short time. So:
1. for job one we pull the latest hd library only and build it.
2. From 2 to 5 we do the same for each availble library.
3. In job 6 we pull the latest of each library and build all of them at the same time.

This way, we're actually testing the build process of the skywater-pdk repo as well. More tests are done on the openlane side, and so we'd be able to catch any break in the loop of dependencies as soon as it happens. 

On that note, would you prefer using a fixed commit that we know works well with open_pdks, or should we always the latest availble version of the repo?

agorararmard:
@RTimothyEdwards: Getting a pre-built version would require having one availble on a public github repo, or purchasing a machine to run the tests on (and installing the pre-built pdk there). Yet, I think that building the pdk each time is an extra test that ensures the sustainability of the whole process. What do you think about that?

RTimothyEdwards:
@agorararmard :  open_pdks doesn't do anything except move some files around and compile some of them into single library files, so I wouldn't expect that it matters what version of skywater-pdk is being used.  It really needs more realistic tests that check if the resulting libraries produce errors when read into magic, qflow, netgen, openlane, etc.

agorararmard:
@RTimothyEdwards: those realistic tests exist on the OpenLANE side (which also uses magic and netgen). The idea of this CI is only to verify that a given Pull Request or Push doesn't break the build. Then comes the rule of the openlane CI that validates the correctness of the produced data, in which case, we would find them on our end and file an issue in either repo. All this [branch](https://github.com/efabless/openlane/tree/develop-latest_pdk) of openlane does, is that has a cron job that clones the latest skywater-pdk and the latest open_pdks, builds them, and tests them with the latest openlane flow against ~30 designs from RTL to GDS-II and compares the results with a group of statistics obtained previously from a certain stable build. Yet, if skywater-pdk or open_pdks build are broken, then this test wouldn't get us anywhere, and so with this open_pdks CI, we'll be one step ahead.

agorararmard:
@RTimothyEdwards: I also have a basic CI setup for netgen and magic, but we could discuss and improve those later (when we're done with these busy weeks, if we're done. :-) )

RTimothyEdwards:
I already merged this, although somehow I failed to mention it here.  Anyway, I will close this issue;  any problems with running the CI itself should be raised in the issue tracker.

PR #44, opened 2020-10-29
Modifications related to the installation of the I/O library

ax3ghazy:
I am installing the custom additions first because otherwise, on my end, the compiled LEF file would get compiled to `library_name.lef` (`sky130_fd_io.lef`) overwriting the pre-existing original LEF.


RTimothyEdwards (approved):
Obviously I didn't read the LEF/DEF documentation and assumed that pads have POWER and GROUND just like pins do. . .  Oh, they have such consistency. . .

RTimothyEdwards:
Pull request has been merged, and will show up in github in about seven hours.

ax3ghazy:
@RTimothyEdwards: It felt so natural that I stupidly went ahead and fixed the OpenROAD parser to read the GROUND token properly only to discover it wasn't standard...

PR #46, opened 2020-10-29
add extension name for techlef option

### LINKED ISSUE #45: Build breaks when not building the sky130_osu_sc_t18 library
The build breaks when the sky130_osu_sc_t18 library is not there, because it tries to move the tlef file with the mv command without checking if the file exists or not, and so the build fails if it's not there (which will always be the case until the library is released).
```
mv `pwd`/sky130A/libs.ref/sky130_osu_sc_t18/techlef/sky130_osu_sc.tlef `pwd`/sky130A/libs.ref/sky130_osu_sc_t18/techlef/sky130_osu_sc_t18.tlef

mv: cannot stat `/home/travis/build/RTimothyEdwards/open_pdks/sky130/sky130A/libs.ref/sky130_osu_sc_t18/techlef/sky130_osu_sc.tlef': No such file or directory
```

The solution to this issue is in this PR #39, but it was closed without being merged :-)

agorararmard:
This along with the recent push to opencircuitdesign open_pdks master fixes #45.

### DIFF
--- a/common/foundry_install.py
+++ b/common/foundry_install.py
@@ -844,6 +844,8 @@ if __name__ == '__main__':
                     fileext = '.cdl'
                 elif option[0] == 'lef':
                     fileext = '.lef'
+                elif option[0] == 'techlef':
+                    fileext = '.tlef'
 
                 if newname:
                     if os.path.splitext(newname)[1] == '':


RTimothyEdwards:
I implemented this in a slightly different way;  the right-hand side of "rename=" should just have the ".tlef" extension.  However, I did note that the part of the code in foundry_install.py does not handle all types and needed a catch-all case at the end as well.

Issue #51, opened 2020-11-04
Remove workaround for missing diodes


mithro:
With https://github.com/google/skywater-pdk/pull/204 being merged, I believe you need to remove the following script https://github.com/RTimothyEdwards/open_pdks/blob/master/sky130/custom/scripts/fix_missing_diodes_lef.py now?

mithro:
@agorararmard

RTimothyEdwards:
This has been updated in the master branch of the opencircuitdesign repo.

RTimothyEdwards:
My script does in fact work for immediate mirrors on demand.  So the github repo has been updated, as has the opencircuitdesign website.

PR #53, opened 2020-11-05
remove the script itself, since it's no longer needed

agorararmard:
This is a follow-up on #51.


RTimothyEdwards:
Merged on opencircuitdesign.com.

PR #63, opened 2020-11-24
Add antenna info for sky130_ef_io pads

ax3ghazy:
- TODO: use the script to also add the antenna info to the sky130_fd_io
  as well


RTimothyEdwards:
Pulled and merged

PR #64, opened 2020-11-24
Update no_synth.cells

ax3ghazy:


RTimothyEdwards:
Pulled and merged

riking:
I've seen several of these giant updates to no_synth.cells with zero explanation - are they automatically generated somehow?

ax3ghazy:
@riking: Yes, I should've left some notes. This is the list of cells, on which I was able to cleanly build mgmt_core in caravel. I have allowed two smaller inverters, and removed some other smaller cells that were causing pin access/congestion issues. The changes aren't that drastic though in this patch. What are your concerns?

They are not automatically generated, but when we face pin access issues in a certain (complex/important) design, we may revise them based on the issues we see.

mithro:
@ax3ghazy -- This very much feels like you are solving the problem in the wrong way here...

riking:
> pin access/congestion issues

This doesn't feel like a valid justification for a global cell banlist - at most, cells causing generic routing congestion should go on a per-project ban list.

ax3ghazy:
I agree; however, most of the time it's more than generic routing congestion. Sometimes the cell is just badly designed, having very tightly-placed or small pins surrounded by obstructions or things like that.

Anyway, the root causes of each of those excludes should be addressed or at least be well-documented. I started a spreadsheet for that purpose at https://docs.google.com/spreadsheets/d/1yujIXMO6jPaVmauqBKc8yL1-ldw_oBhtySXgKSNQ4nM/edit#gid=0 (will also put on several READMEs).

PR #65, opened 2020-11-24
Update openlane pdngen config

ax3ghazy:
- Remove some deprecated settings
- connect VPWR and VPB resp. VGND and VNB to the power/ground nets


RTimothyEdwards:
pulled and merged

PR #69, opened 2020-11-30
Added tri-state buffer mapping file

Manarabdelaty:
Needed to allow yosys to map tri-state buffers.


RTimothyEdwards (approved):
Shouldn't there also be a mapping for the positive-enable tristate buffer?

RTimothyEdwards:
Pulled and merged.

RTimothyEdwards:
Please add the tri-state buffer mapping for all standard cell libraries, not just HD.

Manarabdelaty:
Added in #73

PR #71, opened 2020-11-30
control the global_connections section in common_pdn.tcl with a flag

agorararmard:
This section causes trouble for OpenSTA because it's unable to deal with the power connections when they are too many for some reason. This needs to be opened as an issue on the OpenROAD side. However, let's apply this quick workaround for now since this section isn't really needed for most of the designs.


RTimothyEdwards (approved):
Looks okay.

RTimothyEdwards:
Pulled and merged.

PR #73, opened 2020-12-01
Added tribuff_map to all standard cells

Manarabdelaty:
- Added tribuff_map.v to all standard cells and updated the Makefile.in accordingly
- Fixed ebufn port argument name in hd SCL


RTimothyEdwards:
Pulled and merged.

PR #80, opened 2020-12-19
Changes to netgen setup to enable device level LVS

d-m-bailey:
Before the change, `decap` and `fill` cells were ignored unconditionally.

This change ignores `fill` and `tapvpwrvgnd` cells only if performing device level LVS (`MAGIC_EXT_USE_GDS`).
`fill` and `tapvpwrvgnd` cells are extracted as instances only from GDS (no empty subcircuits).  This causes a placeholder instance to be created with pins that don't watch the netlist spice library.

Ignoring `decap` cells has been removed. There appears to be insignificant performance degradation. Previously, mosfets in the `decap` cells were not extracted, which may have presented a problem. This has been fixed. 

For LEF based LVS, fill cells, `decap` cells, and `tapvpwrvgnd` cells can all be included to verify connections. This may be the simplest way to check that all `fill` cells and `tapvpwrvgnd` cells are connected correctly. Floating `tapvpwrvgnd` cells do not generate errors in device level LVS if at least one is connected correctly. Floating fill cells are also undetectable in device level LVS.


RTimothyEdwards (approved):
I think it is probably better just not to ignore cells for LVS.  Earlier versions of the libraries were giving too much trouble with these.  I think those issues have all been resolved.  I'll merge this change, but it's worth revisiting.

RTimothyEdwards:
Pulled and merged on opencircuitdesign.com.

PR #81, opened 2020-12-22
Restore I/O bus fillers from 0ad7fc9

ax3ghazy:
Those were overwritten by a previous commit (after the "obstructions outside boundaries" fix) and had hand-edited fixes like the removal of some met4 obstructions and adjustments of some met5 shapes.


RTimothyEdwards:
Pulled and merged.

PR #82, opened 2020-12-22
Minor fixes in IO files

ax3ghazy:
- PAD GROUND -> PAD POWER
- Remove leftover pin declarations in a clamped pad


RTimothyEdwards:
Pulled and merged.

PR #85, opened 2020-12-29
Fixes to various views of the IO cells

ax3ghazy:
This mainly removes a 0.015um met5 notch on {,dis}connect cells, which causes min width violations, and fills violating met1 holes in the clamp overlays.


RTimothyEdwards (approved):
Looks good to me.

RTimothyEdwards:
Pulled and merged.

PR #88, opened 2021-01-04
Add a default None value for tclscript

ax3ghazy:
Otherwise, things like https://github.com/RTimothyEdwards/open_pdks/blob/master/common/foundry_install.py#L1235 would be referencing a non-existent variable.


RTimothyEdwards (approved):
Looks good, thanks.

RTimothyEdwards:
Pulled and merged.

PR #91, opened 2021-01-15
Klayout drc

agorararmard:
- Base integration of the klayout drc deck from https://github.com/laurentc2/SKY130_for_KLayout. @ax3ghazy has some batches for that, so he'll be submitting other PRs to fix the .lydrc once this is merged, I believe.
- Updated openlane configs to reflect that change.
- The easiest way to run batch DRC using Klayout is to use the `text` section of the .lydrc file + some adjustments. So, I went ahead and created a script to parse the .lydrc and create a .drc script that could be directly passed to Klayout in batch mode and ran, keeping the old lydrc as is. This happens automatically during the installation of open_pdks so that we'd only keep 1 version of the drc deck, and generate the script from it with each new pdk installation.
- The current .lydrc file contains utf-8 encoded characters which breaks the build if it's passed to the preproc.py script (is it uses ascii encoding). The right way to solve this issue would be to rewrite the file reads and writes in the preproc.py script to use encoding="utf-8". However, I was not sure if this will break something else/produce wrong info in other files. So, I went for the easy solution of copying the file as is (since it doesn't require any processing).


RTimothyEdwards (approved):
Looks okay, I guess.

RTimothyEdwards:
Pulled and merged.

PR #95, opened 2021-01-19
Add missing MINENCLOSEDAREA as patches for now...

ax3ghazy:
Here isn't the proper place to apply this, but this addresses https://github.com/google/skywater-pdk/issues/285 until it gets fixed at  [skywater-pdk](https://github.com/google/skywater-pdk)... In many cases, this causes the only type of DRC errors left in a design, so this is urgently needed. (cc: @mithro)


antonblanchard:
Sounds good. I'm having to apply this manually each time I update my PDK tree right now.

RTimothyEdwards (approved):
Looks good to me.

RTimothyEdwards:
Merged and pulled.

PR #97, opened 2021-01-22
Ignore 236/0

antonblanchard:
During DRC checking of a design I see a lot of:

Error while reading cell "sky130_fd_sc_hd__o21a_4" (byte position 147754): Unknown layer/datatype in boundary, layer=236 type=0

236/0 appears to be an outline layer. Tell magic to ignore it.

### DIFF
--- a/sky130/magic/sky130.tech
+++ b/sky130/magic/sky130.tech
@@ -2226,6 +2226,7 @@ style  sky130 variants (),(vendor)
  ignore POLYMOD
  ignore LOWTAPDENSITY
  ignore FILLOBSPOLY
+ ignore OUTLINE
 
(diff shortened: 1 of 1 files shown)


antonblanchard:
Hi @RTimothyEdwards, I'm trying to cut down on warnings and errors in the openlane flow, but I'm not entirely sure if I'm doing the right thing here.

RTimothyEdwards (approved):
Yes, this is right.

RTimothyEdwards:
Fair enough.  I've waited this long to get an answer on what the undocumented layer 236:0 is supposed to be for, and if I haven't gotten an answer by now, I'm probably not going to get one.

Pulled and merged.

PR #98, opened 2021-01-26
Don't recursively expand SKYWATER_PATH

ax3ghazy:
Otherwise, the following error occurs:
```
Makefile:160: *** Recursive variable `SKYWATER_PATH' references itself
```


RTimothyEdwards:
Pulled and merged, thanks.

PR #99, opened 2021-01-26
sky130_ef_sc_hd__fakediode_2 fixes

ax3ghazy:
This adds the missing HVTP layers in the GDS and eliminates the
"illegal overlaps" caused by the `li1` and `mcon` pin shapes.


RTimothyEdwards:
Pulled and merged.

PR #101, opened 2021-01-27
Fix fake diode name

antonblanchard:
The recent update to the fake diode resulted in a name change, which I think is unintended.

It looks like a couple of commits haven't hit github yet, I can rebase when they do.


antonblanchard:
I'm now seeing:
```
Cell sky130_ef_sc_hd__fakediode_2 couldn't be read
```

Is that because the gds file has the incorrect name:
```
strings ./sky130/custom/sky130_fd_sc_hd/gds/sky130_ef_sc_hd__fakediode_2.gds
sky130_fd_sc_hd
sky130_fd_sc_hd__diode_2
```

ax3ghazy:
Thanks; yes, it was an unintended copy-paste error. https://github.com/RTimothyEdwards/open_pdks/pull/102 fixes both of those issues now.

RTimothyEdwards:
Pulled and merged.

PR #102, opened 2021-01-27
Correct 'fd' -> 'ef' in fakediode

ax3ghazy:
Follow up on #101. Correcting a copy-paste error...


RTimothyEdwards:
Pulled and merged.

PR #103, opened 2021-02-03
Minor openlane config update

agorararmard:
- Reduce default cell padding to 4.
- Add a pointer to no_synth.cells in each library config. This is the first step at aggressively reducing the no_synth list and allowing for a design specific list.
- Copy latch_map.v to the sky130A directory..


RTimothyEdwards:
Pulled and merged.

PR #105, opened 2021-02-08
Fixes and reorganization of the KLayout DRC deck

ax3ghazy:
- define "wildcards" to be used instead of repeating the layer/purpose
  pairs every time
- Move li1 to the back-end layers (FEOL checks are disabled by default)
- Revise li1-related rules,
- Revise all spacing rules for met{1,2,3,4,5} -- use "space" instead of
  "isolated"
- Added a "violators layout" that exercises the reviewed rules


RTimothyEdwards:
Pulled and merged.

PR #107, opened 2021-02-10
Added adder mapping files for yosys

Manarabdelaty:
- to enable $add/$sub operator mapping to different types of adders
- to enable mapping $fa to full adder


RTimothyEdwards:
Pulled and merged.  I also changed the Makefile.in so that all of the openlane install files common to every digital standard cell library are listed in a variable and looped over, instead of having a separate install command for every one.

PR #110, opened 2021-02-11
[openlane] Revised cell exclude lists

agorararmard:
- Made the exclude lists configurable.
- Separated the exclusion lists to no_synth.cells and drc_exclude.cells.
- The merging of the two lists will be used in synthesis.
- Only the drc_exclude.cells will be used in optimizations.
- Aggressively reduced the don't-use list.

More documentation will be added to the openlane side to explain the usage of both.


RTimothyEdwards:
Pulled and merged.

PR #111, opened 2021-02-16
Openlane updates

agorararmard:
-  Map to the smallest mux by default, since resizer always resizes them to the smallest size anyways + mux4_4 makes the latest tritonRoute unhappy, so let's use this as an until-then solution.
- move all probe/c cells to the drc_exclude list


RTimothyEdwards:
Pulled and merged.

PR #112, opened 2021-02-17
OL: Decrease the tapping distance to 13um

ax3ghazy:
- even 14um didn't work in some cases and the claimed fix at
https://github.com/The-OpenROAD-Project/OpenROAD/issues/562 still doesn't do
it, so until we debug and sort it out, we're gonna tap slightly more
often...


RTimothyEdwards:
Pulled and merged.

PR #113, opened 2021-02-22
Update the exclude lists to match the new mapping of muxes to size 1

agorararmard:
Also removing fa_4, as a result of https://github.com/The-OpenROAD-Project/OpenROAD/issues/580


RTimothyEdwards:
Pulled and merged

PR #114, opened 2021-02-23
Fix for older versions of bash

donn:
macOS ships an old version of bash, namely GNU bash, version 3.2.57(1). This is the same shell used for Make. It doesn't support the shorthand &| operator.

This manually expands it to 2>&1 |.


RTimothyEdwards:
Pulled and merged.

PR #115, opened 2021-02-23
Openlane libs updates

agorararmard:
So, we haven't tried the other libs for a while now. This update allows us to use the tap cells in ls and hdll std libs. 

Also, I found that magic complains from `sky130A_fd_sc_hs__decap_8` and `sky130A_fd_sc_ms__decap_8`, and it gives this error `Diffusion contact to gate < 0.055um (licon.11)`. However, I think this is an issue for another day to be reported in the main Skywater pdk repo after @RTimothyEdwards has the time to look into it. For now, let's tell openlane to avoid using those cells.


RTimothyEdwards:
Pulled and merged.

PR #122, opened 2021-05-17
cdl installation scenarios for both EF_STYLE PDK and regular one

ahmednofal:


RTimothyEdwards (approved):
Looks okay;  I will merge it first thing tomorrow because I tend to have troubles with the merge if I don't let my own changes during the day update and get mirrored to github first.  Probably not an issue for a 2-line change, but just to be safe.

RTimothyEdwards:
Pulled and merged.

PR #127, opened 2021-06-13
Rework configure script to be more standard

mithro:
* Use `--prefix` to specify where in the file system to install the PDKs too.

 * Use `DESTDIR` with `make install` to create a distribution that can be extracted to `--prefix` using a package management / system administration system.

 * Continue to support old makefile targets as aliases.

 * Add a `autogen.sh` script.


mithro:
I'm unsure how to continue to support the old `--with-sky130-local-path` / `--with-sky130-dist-path` values....

mithro:
Friendly poke?

RTimothyEdwards (approved):
Looks okay to me.  There are a couple of minor errors but I will patch them up by hand.

RTimothyEdwards:
Pulled, fixed merge conflicts, corrected a few typos, and committed.

Note that for reasons of backwards compatibility, I kept the directory name "pdk", not "pdks".  Also, I made the default prefix to be "/usr", so that by default open_pdks installs into "/usr/share/pdk", not "/usr/local/share/pdk".

Update was made to the repository at opencircuitdesign.com;  the changes will update on github at 3:00am EDT.

mithro:
> Pulled, fixed merge conflicts, corrected a few typos, and committed.
> 
> Note that for reasons of backwards compatibility, I kept the directory name "pdk", not "pdks". Also, I made the default prefix to be "/usr", so that by default open_pdks installs into "/usr/share/pdk", not "/usr/local/share/pdk".

That violates the traditional Linux standard which is;
 * Installs from source should go to `/usr/local`
 * Installs from packages should go to `/usr`

RTimothyEdwards:
Oh, is that a standard?  If so, it has historically been badly conceived and improperly executed, but sure, whatever.

RTimothyEdwards:
Fixed.

mithro:
https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04s09.html

mithro:
See also https://unix.stackexchange.com/questions/11544/what-is-the-difference-between-opt-and-usr-local

mithro:
And https://www.linuxjournal.com/content/filesystem-hierarchy-standard

mithro:
And https://newbedev.com/usr-bin-vs-usr-local-bin-on-linux :-)

PR #134, opened 2021-06-29
fixed bugs with sealring_generator

emayecs:
Fixed bug so that the temp directory is created in the working directory instead of the directory where the script is located. The script also checks if the working directory is writable. Finally, the script also checks if "tmp" is an available directory if "temp" is already taken.

### DIFF
--- a/sky130/custom/scripts/seal_ring_generator/sky130_gen_sealring.py
+++ b/sky130/custom/scripts/seal_ring_generator/sky130_gen_sealring.py
@@ -53,10 +53,12 @@ import re
 
 def generate_sealring(width, height, target_dir, force, keep):
 
+    starting_dir=os.getcwd()
+    
     # Find directory where the script and relevant files exists
     
     abspath = os.path.abspath(__file__)
-    dname = os.path.dirname(abspath)
+    script_dir = os.path.dirname(abspath)
     
     # All files of interest are listed below.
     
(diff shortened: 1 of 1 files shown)


RTimothyEdwards (changes requested):
The only issue is that the fallback directory should be in '/tmp'.  Since that's a global directory used by everyone, the name should be specific to the sealring generator, so maybe something like '/tmp/sky130_sealring".  It is supposed to be a fallback that always succeeds, so the procedure should be (1) check directory "temp" in current working directory;  (2) if that already exists or cannot be created, then print a message saying so, then (3) check directory "/tmp/sky130_sealring".  If that exists, then remove all of its contents.  If it doesn't exist, then create it.  If it cannot be created, then exit with error code 1 (my mistake in giving it a success code of zero) (/tmp by definition is supposed to be writeable, so this shouldn't fail, but you never know).

RTimothyEdwards (approved):
Looks good;  I'll go ahead and merge this.

RTimothyEdwards:
@emayecs :  The way this works is that I do a merge into the repository at opencircuitdesign.com.  The github repository is a mirror of opencircuitdesign.com and updates at 3:00 in the morning.  If you have cloned from github and need an immediate update for any reason, I can force the mirror copy.

Otherwise, this pull request has been pulled and merged.

emayecs:
Sounds good, thanks!

PR #135, opened 2021-07-03
Bug fixes for sky130 scripts

emayecs:
Fixed bugs with setting the correct path for the target files for generate_fill and check_density. Made syntax corrections for check_antenna. Added error catching to all files involving the .rcfiles and checking whether some files existed or not.


RTimothyEdwards (approved):
I didn't go over this with a fine-toothed comb, but the original is so obviously full of errors that this can only be an improvement.  I will run some tests later;  no need to let it gate merging the pull request.

RTimothyEdwards:
Merged and pulled on opencircuitdesign.com;  github mirror will update overnight.

PR #138, opened 2021-07-17
Modified sky130 makefile

emayecs:
The makefile in sky130 now adds the appropriate library import:
`.lib usr/share/pdk/sky130A/libs.tech/ngspice/sky130.lib.spice tt`
into the files in /open_pdks/sources/xschem_sky130/sky130_tests.


RTimothyEdwards (changes requested):
"usr/share/pdk" should be "/usr/share/pdk" (leading slash), but actually both of those are wrong, because it should use ${XSCHEM_STAGING_A}.  Note that the installation script parses every file and replaces the staging area location with the install location.

RTimothyEdwards (approved):
Looks good to me.

RTimothyEdwards:
Make sure you do a "git pull" occasionally to update your version of the repository.  I had to fix a merge conflict in Makefile.in.  Otherwise, the pull request has been merged on opencircuitdesign.com.

PR #139, opened 2021-07-29
 Modified og_gui_manager.py and other scripts

emayecs:
Modified og_gui_manager.py to make it accessible on devices not on the efabless platform. Changed the create project script to make the proper config directories so that the layout and schematic editors can be used. Modified profile.py to make the settings properly reflect the user preferences after saving/closing.


RTimothyEdwards (changes requested):
Please do a git pull, since some modifications appear to pre-date the current version.  Also I just merged pull request #140 which was posted after yours.

RTimothyEdwards (changes requested):
I'm still bothered by a few things.  The main one is that the list of changes shows a change to sky130/Makefile.in which shouldn't be there, because it was already committed (and then edited, because one of the comments ended up in the wrong place).  Possibly the only way to resolve this is for you to remove sky130/Makefile.in and then do a git pull to restore the correct Makefile.in.

Another thing that bothers me is that `create_project.py` has replaced PREFIX with `/usr/share`.  That's wrong, since `/usr/share` is only the default prefix, while PREFIX is substituted with the prefix passed to the configure script (see the top-level Makefile under recipe `common_install`).

RTimothyEdwards (commented):
Of less urgency:  All files  beginning with "og_" should have the name changed.  "og" refers to "open galaxy" and is therefore efabless platform-specific.  "og_config.py" should just be "config.py";  "og_gui_manager.py" should be "project_manager.py", and "og_gui_characterize.py" should probably just be "cace.py".

RTimothyEdwards (approved):
Looks good to me.

RTimothyEdwards:
Pulled and merged (on opencircuitdesign.com)

PR #140, opened 2021-08-01
Update OpenLane config.tcl

Manarabdelaty:
- Added configurations for using a different standard cell library (that is compatible with the base library) when doing resizer optimizations.


RTimothyEdwards (approved):
Looks good to me.

RTimothyEdwards:
Pulled and merged.

PR #141, opened 2021-08-05
project_manager uses info.yaml for basic project information

emayecs:
Replaced project.json with info.yaml in places asking for basic project info. Renamed some variables that still referred to 'og.' Added create_yaml() method. The create_project script creates the yaml file using the yaml loader instead of printing directly into the yaml file.


RTimothyEdwards (changes requested):
Two things:

(1) Why did foundry_nodes.py line 1 change from "/usr/bin/env python3" (correct) to "/usr/bin/python" (incorrect)?

(2) There are a couple of references to Open Galaxy in comments and such using the phrase "Open Galaxy" that should be removed (I see one in profile.py;  there might be others).

emayecs:
Sorry about that Mr. Edwards; I think I copied the original copy of foundry_nodes.py with the old line on accident.

RTimothyEdwards (approved):
Looks good to me.

RTimothyEdwards:
Pulled and merged.

PR #145, opened 2021-08-16
OpenRCX Support 

Manarabdelaty:
Adds openrcx extraction rules file and the related environment configs for openrcx support in openlane


RTimothyEdwards (approved):
Looks good to me.

RTimothyEdwards:
Pulled and merged.

PR #148, opened 2021-08-20
Added flow dialog

emayecs:
The flow button opens the flow dialog which allows the user to choose a flow from the drop down menu, which is saved in the info.yaml file. After that, whenever the flow button is clicked or the project/subproject is double-clicked, the appropriate flow manager is launched. As of now, only the digital flow is able to be used.


RTimothyEdwards (approved):
This looks okay---since nothing is gated by it, I'll go ahead and merge it and test in situ.

RTimothyEdwards:
Pulled and merged.

PR #149, opened 2021-08-21
Changed qflow install recipe so setup file names match pdk node names.

emayecs:
Also changed the synthesize method in project manager so the pdk qflow directory is given to qflow_manager.py instead of just the pdk directory.


RTimothyEdwards (approved):
Looks okay to me.

RTimothyEdwards:
Pulled and merged.

PR #150, opened 2021-08-23
Bug Fixes

emayecs:
- Project Manager used an incorrect path to the qflow directory, causing the technologies to not show up in the Qflow Manager
- sky130 Makefile was copying from a qflow setup file with the wrong name

### DIFF
--- a/common/project_manager.py
+++ b/common/project_manager.py
@@ -3851,7 +3851,7 @@ class ProjectManager(ttk.Frame):
         if value:
             design = value['values'][0] # project path
             pdkdir = self.get_pdk_dir(design, path = True)
-            qflowdir = pdkdir + 'libs.tech/qflow'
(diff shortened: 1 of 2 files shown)


RTimothyEdwards (changes requested):
I didn't notice before, but there are also ".par" files in addition to the ".sh" files.  The ".par" files are not called out within the ".sh" file, so I assume that I set it up such that the ".par" file should have the same root name as the ".sh" file, and will likely cause a failure if it doesn't.

RTimothyEdwards (commented):
Not to mention that also in the Makefile.in file there are "rm -f" statements that clean up the ".sh" and ".par" files, and those names need to be changed as well.

RTimothyEdwards (approved):
Looks good to me.

RTimothyEdwards:
Pulled and merged.

PR #153, opened 2021-08-28
Repo structure changes

emayecs:
- Modified shebang line at the top of certain python files
- Moved project manager-oriented scripts from common folder to runtime folder
- Makefile.in installs scripts into datadir/pdk/scripts
- Config file installs natsort
- Added help .txt files to runtime folder


RTimothyEdwards (changes requested):
Umm. . . almost but not quite.  The "sort_pdkfiles.py" script is used by the make/install process and so must remain in common/.

There are several other scripts that I think I just left in open_pdks because I didn't have anywhere else to put them;  those are called from neither the run-time applications or from the make/install process (they are part of the make/install process for skywater-pdk;  they got modified there and I wanted to keep the original versions which is why they are in open_pdks).  They should also remain in common/:  compare_dirs.py, find_all_devices.py, fix_subckt_params.py, fixspice.py, get_gds_date.py, print_subckt_params.py, remove_specify.py, spectre_to_spice.py, split_gds.py, split_one_spice.py, and split_spice.py (I checked that these are not called either from sky130/Makefile.in or from any other script in common/, but you probably should double-check).

RTimothyEdwards (approved):
The change to "configure" is ineffectual---it will get overwritten by the next run of "autoconf".  Changes like that need to be made to "configure.ac".  However, "make" should be altering the environment, not "configure".  "pip install" will normally need to be run "sudo" and so would just raise an error if part of "configure".  I think that "configure" should just be checking for "natsort", which may require a bit of custom scripting in the configure.ac file, and simply report a failure if it is not present and indicate to the user that it needs to be installed.

Also, I think "requirements.txt" is related to the python virtual environment.  Looking at the contents, though, it's possible that the configure script needs to check for "matplotlib" as well as "natsort".  By the way, both "matplotlib" and "natsort" are run-time requirements, not build requirements, so the build should keep going whether or not they exist.

Issue #59, opened 2020-11-18
generic resistors don't match in LVS


d-m-bailey:
The `open_pdks/sky130/magic/sky130.tech` file extracts generic resistors as R-devices.

Generic resistors
```
sky130_fd_pr__res_generic_l1
sky130_fd_pr__res_generic_m1
sky130_fd_pr__res_generic_m2
sky130_fd_pr__res_generic_m3
sky130_fd_pr__res_generic_m4
sky130_fd_pr__res_generic_m5
sky130_fd_pr__res_generic_nd
sky130_fd_pr__res_generic_nd__hv
sky130_fd_pr__res_generic_pd
sky130_fd_pr__res_generic_pd__hv
sky130_fd_pr__res_generic_po
```
Sample extraction (2 terminal resistor):
```
.subckt sky130_fd_sc_hd__conb_1 VGND VNB VPB VPWR HI LO
R0 HI VPWR sky130_fd_pr__res_generic_po w=480000u l=45000u
R1 VGND LO sky130_fd_pr__res_generic_po w=480000u l=45000u
.ends
```
The spice library has the 3 terminal X-resistor definitions.
```
.subckt sky130_fd_sc_hd__conb_1 VGND VNB VPB VPWR HI LO
X0 VGND LO VNB short w=480000u l=45000u
X1 HI VPWR VNB short w=480000u l=45000u
.ends
```
Maybe change the extraction to `rsubcircuits` and change the spice library to 2 terminal `sky130_fd_pr__res_generic_po` devices.

Also the variant(orig) sytle has a spelling error.
```
 device rsubcircuit sky130_fd_pr_res_generic_nd     ndiffres \
        *ndiff pwell,space/w  error l=l w=w
 device rsubcircuit sky130_fd_pr_res_generic_pd     pdiffres \
        *pdiff nwell    error l=l w=w
```

Probably should be `sky130_fd_pr__res_generic_nd` and `sky130_fd_pr__res_generic_pd` (2 '_' between the 'pr' and 'res')

RTimothyEdwards:
The spelling errors have been noted and I am fixing them now.

The resistor types takes some working through, because the models are in different places and some really are R type devices.  The changing PDK was a moving target for a while.  The "short" resistor is different from the others, and I have been adding that manually to the SPICE library because it is not part of the PDK.  The "rsubcircuit" device can take a substrate node, so it can generate 3-terminal devices in the netlist.

[...]

PR #154, opened 2021-08-31
Corrects 3 terminal x-device resistor conversion.

d-m-bailey:
Changes 3rd terminal to $SUB=xxx where xxx is the 3rd terminal.

Should resolve the remaining issues with #59.

For example, the change to `sky130_fd_sc_hd.spice` would be

Before
```
.subckt sky130_fd_sc_hd__conb_1 VGND VNB VPB VPWR HI LO
R0 VGND LO VNB sky130_fd_pr__res_generic_po w=480000u l=45000u
R1 HI VPWR VNB sky130_fd_pr__res_generic_po w=480000u l=45000u
.ends
```

After
```
.subckt sky130_fd_sc_hd__conb_1 VGND VNB VPB VPWR HI LO
R0 VGND LO  sky130_fd_pr__res_generic_po w=480000u l=45000u $SUB=VNB
R1 HI VPWR  sky130_fd_pr__res_generic_po w=480000u l=45000u $SUB=VNB
.ends
```

Unfortunately, this results in the following LVS property errors. These can be ignored via a netgen setup change which I'll submit later.
```
There were property errors.
sky130_fd_pr__res_generic_po1 vs. sky130_fd_pr__res_generic_po0:
Property $SUB in circuit2 has no matching property in circuit1
sky130_fd_pr__res_generic_po0 vs. sky130_fd_pr__res_generic_po1:
Property $SUB in circuit2 has no matching property in circuit1
```

### DIFF
--- a/sky130/custom/scripts/fix_device_models.py
+++ b/sky130/custom/scripts/fix_device_models.py
@@ -32,7 +32,7 @@ def filter(inname, outname):
 
     dioderex = re.compile('.*[ \t]+sky130_fd_pr__diode_pw2nd[ \t]+')
     ndioderex = re.compile('.*[ \t]+ndiode_h[ \t]+')
-    shortrex = re.compile('.*[ \t]+short[ \t]+')
+    shortrex = re.compile('(.*[ \t]+)(\S+)([ \t]+short[ \t]+.*)')
 
     for line in slines:
 
@@ -57,8 +57,9 @@ def filter(inname, outname):
             fixedlines.append(fline)
             modified = True
         elif smatch:
-            fline = re.sub('short', 'sky130_fd_pr__res_generic_po', line)
-            fline = re.sub('^X', 'R', fline)
+            fline = ( re.sub('^X', 'R', smatch.group(1)) 
+                    + re.sub('short', 'sky130_fd_pr__res_generic_po', smatch.group(3)) 
+                    + " $SUB=" + smatch.group(2) )
             fixedlines.append(fline)
             modified = True
         else:


RTimothyEdwards:
I cannot find any behavior of "$SUB" documented in ngspice.

I would think that the "usual" way to handle this would be to change the "R" type device to an "X" type device, and then implement the subcircuit as a combination of two resistors of 1/2 length and a capacitor to the substrate (resistor tee).  That would automatically be compatible with all SPICE implementations.

RTimothyEdwards:
Obviously my memory is faulty.  There is no such fix in the patches or scripts.  But it is easy to implement in the `fix_device_models.py` script.

d-m-bailey:
@RTimothyEdwards Thanks for taking the time to look into this. The `$SUB=` parameter appears to be a Calibre extension to spice. Since it's coded as a comment ngspice would hopefully ignore it. In Calibre LVS it's used to check the underlying substrate without having to extract an extra device. CVC can handle the connection similarly, for example, checking for forward biased parasitic diodes in diffusion resistors. Since the device in question is a poly resistor, there is no need for CVC to check the SUB connection, so it won't make a difference if it is just deleted.

RTimothyEdwards:
ngspice documentation specifically states that '$' is only interpreted as a comment character if it is followed by whitespace.  So '$SUB' will likely throw an error in ngspice (I haven't tested it).

RTimothyEdwards:
I pushed a fix to open_pdks that removes the VNB connection from the poly resistors in the conb cells of all the digital standard cell libraries, so I'll take it from your comments above that that's a sufficient solution.

PR #157, opened 2021-09-02
Modified top-level makefile common_install script

emayecs:
Modified makefile to install runtime files using a for loop, script files refer to PREFIX correctly


RTimothyEdwards (approved):
I'll merge this, because it does work, although something simpler would be preferred if you can figure out a more direct way to list just the base name of all the files.

RTimothyEdwards:
Pulled and merged.

PR #160, opened 2021-09-14
Add DATA_WIRE_RC_LAYER, CLOCK_WIRE_RC_LAYER to config

donn:
As per discussion held in https://github.com/The-OpenROAD-Project/OpenLane/pull/605, these values are too PDK-specific to be in OpenLane.


RTimothyEdwards (approved):
Looks good to me.

RTimothyEdwards:
Pulled and merged.

PR #164, opened 2021-10-04
Add default max transition for synthesis and CTS

Manarabdelaty:
- Adds a default maximum transition value for openlane (750 ps)


RTimothyEdwards (approved):
Looks good to me.

RTimothyEdwards:
Pullled and merged.

PR #170, opened 2021-10-26
Patch wire load model to the liberty files

Manarabdelaty:
- Adds a simple wire load model to the hd liberty file so that abc can take some parasitic information into account.


RTimothyEdwards:
I pulled the added parts of the liberty file out of your patch file and dropped them into a separate file called "wireload.lib", and wrote a script that inserts it into each liberty file.  Please advise as to whether there might be a more specific wireload model for other corners than tt or for other standard cell libraries than hd.  The script can be modified to accommodate additional models if you have them.

RTimothyEdwards:
@Manarabdelaty :  Please check the updated repository (commit 7ea416610339d3c29af9d0d748ceadd3fd368608).

RTimothyEdwards:
I'll assume that the script I wrote is working correctly and will close this pull request.

PR #171, opened 2021-11-08
Allow SHARED_PDKS_PATH to be set externally.

mguthaus:
Only set SHARED_PDKS_PATH if it is not already set externally.


RTimothyEdwards (approved):
Looks good to me.

RTimothyEdwards:
Pulled and merged.

PR #172, opened 2021-11-13
Replace SYNTH_DRIVING_CELL with size 2 inverters

donn:
As it stands, the size 1 inverters are the SYNTH_DRIVING_CELL for all SCLs. Problem is, they're also on the no synth list.

See https://github.com/The-OpenROAD-Project/OpenLane/issues/692


RTimothyEdwards (approved):
Looks good to me.

RTimothyEdwards:
Pulled and merged.

PR #181, opened 2021-11-22
Remove timestamp from .mag output files.

### LINKED ISSUE #179: Generated .mag files seem to be non-deterministic?
The output from the auto-builder seems to indicate that the .mag files generated from open_pdk are non-deterministic (IE they change even given identical input).

I'm guessing that the `.mag` file has some type of timestamp in them? I seem to recall something similar from the `skywater-pdk` generation process?

mithro:
Fixes #179.

Signed-off-by: Tim 'mithro' Ansell <me@mith.ro>

### DIFF
--- a/common/foundry_install.py
+++ b/common/foundry_install.py
@@ -1917,7 +1917,9 @@ if __name__ == '__main__':
                                         print('string ' + prop, file=ofile)
                                 for prop in prop_lines:
                                     print('string ' + prop, file=ofile)
-
+                            # Clear timestamp in `.mag` files.
+                            if line.startswith('timestamp '):
+                                line = 'timestamp 0'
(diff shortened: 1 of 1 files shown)


RTimothyEdwards:
The timestamps are the only form of version control in magic, so removing them effectively removes version control.  My plan is to replace timestamping with a checksum, such that magic no longer cares when a file was created, but what the contents are.  I have not had the time to implement that yet.

mithro:
@RTimothyEdwards - I'm unsure why Magic should care about the creation time of files that it isn't modifying (IE these files are installed read-only)?

RTimothyEdwards:
@mithro :  That is likely a bug in magic.

PR #186, opened 2021-11-23
Use `pipefail` so make sees command failures.

### LINKED ISSUE #180: foundry_install.py script failing doesn't stop the build
I was making some changes to `foundry_install.py` script and notice that even when I break it totally, it doesn't seem to stop the `make` script?
```
Annotating files in /home/tansell/github/RTimothyEdwards/open_pdks/sky130/sky130A/libs.ref/sky130_fd_pr/maglef
Traceback (most recent call last):
  File "/home/tansell/github/RTimothyEdwards/open_pdks/sky130/../common/foundry_install.py", line 1922, in <module>
    print("Cleared timestamp in", malefgfile)
NameError: name 'malefgfile' is not defined
```

mithro:
In a pipeline, normally only the exit code of the last command matters.
Thus when sending stuff through tee, a failure of the first command is lost. `pipefail` makes sure that a failure on the first command causes a failure exit code.

More info at
https://stackoverflow.com/questions/6871859/piping-command-output-to-tee-but-also-save-exit-code-of-command

Fixes #180 - "foundry_install.py script failing doesn't stop the build"

### DIFF
--- a/sky130/Makefile.in
+++ b/sky130/Makefile.in
@@ -129,7 +129,12 @@
 #
 #--------------------------------------------------------------------
 # This Makefile contains bash-isms
-SHELL = bash
+SHELL := /bin/bash
+export SHELL
+# We use pipes to save output to files, without pipefail they will always be
+# seen by make as having succeeded.
+SHELLOPTS := pipefail
+export SHELLOPTS
 MV = mv
 SED = @SED@
 


RTimothyEdwards (approved):
Looks good to me.

RTimothyEdwards:
Pulled and merged.

PR #194, opened 2021-11-27
Updated Klayout DRC MR deck and klayout tech file

marwaneltoukhy:
1. Updated the klayout DRC MR deck to the latest stable version (same as caravel's precheck deck)
2. Changed DRC deck from `sky130A_mr.lydrc` to `sky130A_mr.drc` so that its executable without any extra scripts
3. removed `gen_run_drc.py` as we don't need it anymore and changed Makefile accordingly
4. Updated `sky130.lyt` with better options for importing def files with the right layer mapping (there is a bug in klayout that maps special routing and normal routing to the same variable, [we filed an issue on klayout github](https://github.com/KLayout/klayout/issues/942#issue-1063560528))


RTimothyEdwards (approved):
Looks good to me.

RTimothyEdwards:
Pulled and merged.

PR #196, opened 2021-11-29
Improve the CI run times by using less aggressive compression settings

### LINKED ISSUE #193: The "all" build takes almost 2 hours due to capturing
From https://github.com/RTimothyEdwards/open_pdks/runs/4327353828?check_suite_focus=true 

The run `succeeded 2 hours ago in 1h 47m 35s` of which `1h 16m 53s` was used to create the tarball while only `29m 12s` was used in the building.

This is significantly longer than the other builds which average ~30 minutes;
Screenshot from 2021-11-25 14-07-55

mithro:
The `Run (all)` job previously took ~90 minutes, it now takes ~45 minutes. Most of the other jobs also reduce by 50% too. This fixes #193 .

The negative is that it doubles the output artifact size.

| Before | **After** |
| --------- | ------ |
| Screenshot from 2021-11-28 09-57-35 | Screenshot from 2021-11-28 18-45-08 |
| --------- | ------ |
| Screenshot from 2021-11-28 18-46-39 | Screenshot from 2021-11-28 18-46-46 |

### DIFF
--- a/.github/capture.sh
+++ b/.github/capture.sh
@@ -40,7 +40,7 @@ cp .github/magic.tar.gz ${GITHUB_WORKSPACE}/output/
 	cd ${SKY130_DIR}
 	tar \
 		--create \
-		--xz \
+		--gzip \
 		--verbose \
 		\
 		--mtime='2020-05-07 00:00Z' \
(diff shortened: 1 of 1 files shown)


mithro:
@RTimothyEdwards - The speed up in run time seems like a better trade off than the reduced artifact size.

RTimothyEdwards (approved):
Looks good to me.

RTimothyEdwards:
Pulled and merged.

PR #213, opened 2022-01-31
Update config.tcl

marwaneltoukhy:
update klayout paths


RTimothyEdwards (changes requested):
That is not a correct path.  PDK_ROOT should refer to the path above the PDK name, so if the PDK is installed at `/usr/share/pdk/sky130A`, then PDK_ROOT would be `/usr/share/pdk`.

Your path appears to point to the open_pdks source, not the staging area or installation destination, so that is an invalid path to use.

PR #220, opened 2022-02-17
2022.2.17_01.04 release KMR DRC

marwaneltoukhy:
new KMR DRC to exclude SRAM from the new rules in the last release


RTimothyEdwards (commented):
So _any_ cell containing the letters "sram" in the name is exempt from these rule checks?  There are only two layouts that have the errors that we want to ignore, and they should be called out specifically by name.

But I don't see why this belongs in open_pdks.  It's a one-off hack to solve a problem of the moment.  There _is_ an error in two of the SRAM macros, and the choice not to flag them lies solely with the efabless precheck.

PR #225, opened 2022-02-25
Klayout tech files and reram DRC deck

marwaneltoukhy:
- Added tech files for klayout
  - lyp file
  - klayoutrc file
- Added reram DRC deck


RTimothyEdwards (commented):
These files need to start using the preprocessor directives to distinguish between variants A and B.  ReRAM additions, for example, should be inside a block  `#ifdef RERAM` ... `#endif`.  Likewise for redistribution layers (`#ifdef REDISTRIBUTION`), metals 4 and 5 (`#ifdef METAL5`, although this is optional as there does not appear to be any intent by SkyWater to redo the 3-metal stack in sky130), and MiM caps (`#ifdef MIM`, also optional as the MiM caps appear to be a permanent part of sky130).

RTimothyEdwards:
So there should not be an additional file for the MR DRC deck for ReRAM, just add the additional rules or changes to the existing file, inside the `#ifdef` block.  Specific test violators for ReRAM should be split into a separate GDS file.

PR #227, opened 2022-02-28
Updated /sky130/readme to include reram library path

3x10e8:
Added reram path as an optional git submodule for completeness. Might need corresponding update to the documentation on http://www.opencircuitdesign.com/open_pdks/


RTimothyEdwards:
This isn't urgent because I'm not using anything from the reram repository, although at some point I have to deal with the verilog-A device model.  I'll get around to merging it in a few days.  If I haven't done it by Thursday, please send me a gentle reminder.

RTimothyEdwards:
Pulled and merged.

PR #240, opened 2022-03-23
Layout prefix 

d-m-bailey:
Enable ignoring, matching, and parallel reduction for layout cells that have 2 character prefix(s) added.
Layout cells are assumed to be circuit1.

Warning: this commit includes modifications that have not been tested.
I modified the sky130A_setup.tcl in the pdk and tested matching.
Ignoring cells and parallel reduction have not been tested.


d-m-bailey:
Deleted the repo to make a clean copy.  I'll resubmit these changes later.

PR #248, opened 2022-04-05
Fix default_fanout_load

antonblanchard:
Some of the sky130 liberty files have default_fanout_load set to 1,
others have it set to 0. When set to 0 all fan out management and
violation reporting is disabled, so change any instances of 0 to 1.


RTimothyEdwards (approved):
Great addition, thanks!  Also thanks for finding the "break" statements that prevent filters from being chained together like they are supposed to be able to do.

RTimothyEdwards:
Pulled and merged (in the opencircuitdesign.com repository).

antonblanchard:
Thanks @RTimothyEdwards!

PR #249, opened 2022-04-06
Fix unterminated #ifdef in magic/sky130.tech

antonblanchard:


antonblanchard:
@RTimothyEdwards I'm guessing this is where it should go, but please double check I got it right.

RTimothyEdwards (approved):
Apparently my preprocessor script misses underminated ifdefs?

RTimothyEdwards:
Pulled and merged.

antonblanchard:
> Apparently my preprocessor script misses underminated ifdefs?

It does complain, but I guess it doesn't cause the build to error out. I noticed it because I was redirecting stdout to a file and I got a couple of warnings to stderr.

antonblanchard:
Closing

PR #250, opened 2022-04-06
Remove Various Deprecated Variables

donn:
- Remove `VIAS_RC`, `LAYERS_RC`: That information is now in the tech lef
- Remove `GLB_RT_L1_ADJUSTMENT`: Variable has been deprecated forever


RTimothyEdwards (approved):
Looks good to me.

RTimothyEdwards:
Pulled and merged, on opencircuitdesign.com

donn:
Thanks, mind syncing to GitHub?

antonblanchard:
I'm not sure about the change to remove `LAYERS_RC` and `VIAS_RC`. Aren't they meant to be estimates based on feedback from routed designs? Does the data in the TLEF take this into consideration? @maliberty FYI

I think @jjcherry56 has a script to create these estimates.

jjcherry56:
The correlate_rc target in ORFS/flow/util/util.mk is used to correlate the cap values (nothing for res in there).
Not exactly portable to LEF (completely different units).
OR will use the TLEF values; the set_layer_rc command is mostly to override the TLEF when it is broken (most of the time).
set_layer_rc allows corner specific values (TLEF does not).

PR #259, opened 2022-06-30
add n-well generator

pepijndevos:


RTimothyEdwards (approved):
Now that you're an expert on writing device generator scripts in Tcl for magic, I've got a lot of jobs for you. . . !

RTimothyEdwards:
I do see that you just copied code from the deep nwell drawing routine. . .  And didn't even change the error text.  : )   I will correct for the main issue, which is the minimum dimension requirement which is based on deep nwell dimensions and is much larger than the requirement for a simple nwell.

PR #260, opened 2022-07-04
Fix a few places we modify the techlef and don't mark it modified

antonblanchard:
Note: I don't think this results in an issue because we make other modifications to the file, which sets modified.


RTimothyEdwards (approved):
Looks good, will merge.

RTimothyEdwards:
Pulled and merged.

PR #261, opened 2022-07-04
Regenerate configure

antonblanchard:


RTimothyEdwards (commented):
Maybe `configure` itself should just be removed from the repository and require the user to run autoconf, but maybe I can just pay more attention so that I don't commit my local configuration, which is not going to be very useful to people.

antonblanchard:
Sounds good, closing

PR #267, opened 2022-07-29
Config Updates + Magic Update

donn:
+ Add CVC files to PDK proper instead of being inside OpenLane
~ sky130.json whitespace reformat
~ sky130:magic -> `f7df5e7c86fb47c5fd445c846afddc6fbabad6ae`
~ `CELL_PAD` -> `GPL/DPL_CELL_PADDING`
~ `CELL_PAD_EXECLUDE` -> `CELL_PAD_EXCLUDE`
~ Update Klayout file locations


RTimothyEdwards (approved):
Looks okay to me.

RTimothyEdwards:
@donn :  Pulled and merged

My only quibble is that "cvc" is a standalone tool and therefore in principle should have its own base directory and installation recipe instead of being a subdirectory of "openlane".

PR #271, opened 2022-08-16
Added missing metal resistor symbols for Xschem

wulffern:
I'm using some metal resistors in local interconnect and metal 3, and the symbols were missing. To get the LVS correct I copied the res_generic_m1 to the other metal layers.

Example files: 
Sch: https://github.com/wulffern/aicex/blob/main/ip/sun_pll_sky130nm/design/SUN_PLL_SKY130NM/CAP_LPF.sch
Lay: https://github.com/wulffern/aicex/blob/main/ip/sun_pll_sky130nm/design/SUN_PLL_SKY130NM/CAP_LPF.mag


wulffern:
Never mind, I thought the symbols was in git, but that's obviously not the case

PR #272, opened 2022-08-23
sky130B missing THICKNESS LEF props, breaking antenna repair

antonblanchard:
On a recent tape out of Microwatt I noticed an absence of any
antenna diodes as well as no antenna violations reported at the signoff
step.

It turns out we are stripping the THICKNESS properties from the sky130B
tech LEF, meaning all side area calculations come out as zero.

While fixing this, I also noticed filecmp.cmp only checks stat values
by default. It incorrectly determines that the sky130A and sky130B tech
LEFs are identical because the sizes and timestamps match, and creates
a symlink. Use shallow=False to force an actual comparison of the
contents.


RTimothyEdwards (approved):
Thanks.  I had no idea that `filecmp.cmp` was not doing a complete comparison.  Ugh.

RTimothyEdwards:
Pulled and merged.

PR #275, opened 2022-09-03
Propagate subcut and dnwell up the hierarchy

### LINKED ISSUE #276: subcut has to be added to all parent hierarchies
At least that's what I heard.

PR #275 remedies this.

d-m-bailey:
Fixes #276

### DIFF
--- a/sky130/magic/sky130.tech
+++ b/sky130/magic/sky130.tech
@@ -2460,10 +2460,17 @@ style  sky130 variants (vendor),()
  variants *
  labels SUBPIN port
 
- layer dnwell DNWELL
+ templayer alldnwell DNWELL,lowerdnwell
+ copyup lowerdnwell
+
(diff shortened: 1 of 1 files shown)


RTimothyEdwards (approved):
Looks good to me.

RTimothyEdwards:
Pulled and merged.

RTimothyEdwards:
I reverted this, because it does something that I _thought_ I had fixed in magic;  that is, magic should treat deep nwell and isosub as propagated up the hierarchy automatically when doing extraction.  That was a recent (about a year ago?) fix in magic and I don't know if this modification predates that fix and was a workaround.  Regardless, the underlying problem should be handled by magic internally and not by the cifinput rules.  Is there a simple example that can be used to determine if this problem still exists or not in magic (hopefully simpler than extracting the whole Caravel chip)?

PR #284, opened 2022-09-15
update magic commit id for gf180

kareefardi:


RTimothyEdwards (approved):
Also looks good to me.

RTimothyEdwards:
Pulled and merged.

PR #292, opened 2022-10-09
Use calibre calibration for sky130A

donn:


RTimothyEdwards (approved):
Uses the Calibre-derived rules if they exist, so it works around the lack of Calibre-derived rules for sky130B, but will automatically make use of them whenever they become available.  Good implementation.

RTimothyEdwards:
@donn :  Pulled, merged, and mirrored.

Issue #293, opened 2022-10-18
DRC failing MIM Capacitor Device Generator


mtseng15:
# Issue
- Device generator for mim caps creates DRC failing capacitors for both M3 and M4 caps. Please see screen shots below.

<img width="680" alt="Screen Shot 2022-10-17 at 7 07 53 PM" src="https://user-images.githubusercontent.com/49569515/196309087-60b0874e-a5ec-4bbf-8bab-ebb7fed0c8ef.png">
<img width="987" alt="Screen Shot 2022-10-17 at 7 07 35 PM" src="https://user-images.githubusercontent.com/49569515/196309116-370e0a2f-bb4b-46ea-bf90-994859e30b8f.png">

# Desired Behavor
- Device generator for mim caps produces DRC clean capacitors.

mtseng15:
Fixed with merge of pull request above.

PR #294, opened 2022-10-18
Adjusting default mim cap generator values to create DRC clean caps

mtseng15:
Ref: #293 

Adjusted default mim cap device generator values to create DRC clean capacitors (both Magic with DRC full and Klayout). Please see new screen shots below.

Screen Shot 2022-10-17 at 7 27 25 PM
Screen Shot 2022-10-17 at 7 27 05 PM


mtseng15:
Ok, now this passes both Klayout and Magic DRC. I had a mistake in my Klayout script so it wasn't checking everything previously and now it is.

RTimothyEdwards (approved):
Looks good to me.  Sorry about the delay.

PR #304, opened 2022-11-02
fixed klayout tech files and DRC deck

marwaneltoukhy:
Added correct tech files for klayout
Added DRC deck that is used in precheck to tapeout caravel


RTimothyEdwards (changes requested):
But I'm already installing the sky130_mr.drc file _from_ precheck, so that it's not necessary to maintain a redundant copy of the file in open_pdks.  Are you using the file installed in, e.g., `libs.tech/klayout/drc/sky130A_mr.drc`?

RTimothyEdwards (commented):
Plus your changes to the sky130.lyp file look like the changes that are already in, e.g., `libs.tech/klayout/tech/sky130A.lyp`.

PR #311, opened 2022-11-23
Add calibrated OpenRCX rule decks/update openlane configurations

passant5:

PR #312, opened 2022-11-24
gf180 PVT corner and exclude list update

vijayank88:
Signed-off-by: vijayank88 <paruthi143@gmail.com>

Based on ORFS [config.mk](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/master/flow/platforms/gf180/config.mk) update OpenLane flow `config.tcl`.

Also updated `drc_exclude.cells` and `no_synth.cells` to remove `*_1` cells based on https://github.com/The-OpenROAD-Project/OpenLane/issues/1493.

cc/@RTimothyEdwards


RTimothyEdwards:
@vijayank88 :  I am waiting on the openlane developers to check, review, and approve this pull request.

RTimothyEdwards:
@marwaneltoukhy :  I need a review of this from you, but I could not add you as a reviewer;  I think you need to be listed as a collaborator in the repository.  I made an invitation for you as a collaborator.  I think I need you to accept the invitation before I can add you as a reviewer.

gf180mcu/openlane/config.tcl
```diff
@@ -15,9 +15,9 @@ if { ![info exist ::env(STD_CELL_LIBRARY_OPT)] } {
 }
 
 # Lib Files
-set ::env(LIB_SYNTH) "$::env(PDK_ROOT)/$::env(PDK)/libs.ref/$::env(STD_CELL_LIBRARY)/liberty/$::env(STD_CELL_LIBRARY)__tt_025C_3v30.lib"
+set ::env(LIB_SYNTH) "$::env(PDK_ROOT)/$::env(PDK)/libs.ref/$::env(STD_CELL_LIBRARY)/liberty/$::env(STD_CELL_LIBRARY)__tt_025C_5v00.lib"
 set ::env(LIB_FASTEST) "$::env(PDK_ROOT)/$::env(PDK)/libs.ref/$::env(STD_CELL_LIBRARY)/liberty/$::env(STD_CELL_LIBRARY)__ff_n40C_5v50.lib"
-set ::env(LIB_SLOWEST) "$::env(PDK_ROOT)/$::env(PDK)/libs.ref/$::env(STD_CELL_LIBRARY)/liberty/$::env(STD_CELL_LIBRARY)__ss_125C_1v62.lib"
+set ::env(LIB_SLOWEST) "$::env(PDK_ROOT)/$::env(PDK)/libs.ref/$::env(STD_CELL_LIBRARY)/liberty/$::env(STD_CELL_LIBRARY)__ss_125C_4v50.lib"
```
marwaneltoukhy, line 20: This has already been updated in #311

gf180mcu/openlane/gf180mcu_fd_sc_mcu7t5v0/config.tcl
```diff
@@ -11,14 +11,14 @@ set ::env(FP_WELLTAP_CELL) "$::env(STD_CELL_LIBRARY)__filltie"
...
-set ::env(SYNTH_CLK_DRIVING_CELL_PIN) "ZN"
+set ::env(SYNTH_DRIVING_CELL) "$::env(STD_CELL_LIBRARY)__buf_4"
+set ::env(SYNTH_DRIVING_CELL_PIN) "Z"
+set ::env(SYNTH_CLK_DRIVING_CELL) "$::env(STD_CELL_LIBRARY)__buf_4"
+set ::env(SYNTH_CLK_DRIVING_CELL_PIN) "Z"
 
 # update these
-set ::env(SYNTH_CAP_LOAD) "72.91" ; # femtofarad from pin I in liberty file
-set ::env(SYNTH_MIN_BUF_PORT) "$::env(STD_CELL_LIBRARY)__buf_1 I Z"
+set ::env(SYNTH_CAP_LOAD) "9.096" ; # femtofarad from pin I in liberty file
```
marwaneltoukhy, line 20: Not sure where you got these values from, seems like it's too low. Can you please point me to how you got this value?

marwaneltoukhy (changes requested):

mo-hosni (commented):
From my experience with GF180, excluding the size 1 cells increase the area by like 20%.

I don't know if that should be the default. I mean if someone is working with a clock period of 40ns this probably won't do him any favor. Also, the small macros won't benefit from this list, this will cause an increase in area without any advantage in timing for such macros.

I actually always exclude those cells when timing is a priority, and I also exclude size 2 cells. But, when area is the priority, I don't use such lists.

So,
I think a better solution is to use different lists based on the synthesis strategy.  Or leave it for the designers to make their own lists instead of making a default that prioritize timing over area.

gf180mcu/openlane/gf180mcu_fd_sc_mcu7t5v0/config.tcl
vijayank88: As per ORFS [config.mk](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/710a6af4e2934ad968e3967e5ae99707206e7e1a/flow/platforms/gf180/config.mk#L50) for 9track `export ABC_LOAD_IN_FF                         = 0.01343` 
For 7track *_buf_4 `capacitance : 0.009315 ;` . Its should be `9.315` as OL flow converting to FF

gf180mcu/openlane/config.tcl
vijayank88: fixed

gf180mcu/openlane/gf180mcu_fd_sc_mcu7t5v0/config.tcl
marwaneltoukhy: 1. The values in ORFS is different
2. Where did you get the value in ORFS?

vijayank88: `libs.ref/gf180mcu_fd_sc_mcu9t5v0/liberty/gf180mcu_fd_sc_mcu9t5v0__tt_025C_5v00.lib`

kareefardi:
@vijayank88 can you comment on the feedback from @mo-hosni

vijayank88:
Yes as per @mo-hosni  statement its correct.
Most of designs are timing critical. So I try to avoid `_1` cells.
Based on sky130 setup, i've created this list for timing critical designs.
So its upto designers to decide which cells to use/avoid. 
I leave it to experts approval to merge or skip. Thanks

antonblanchard:
> Yes as per @mo-hosni statement its correct. Most of designs are timing critical. So I try to avoid `_1` cells. Based on sky130 setup, i've created this list for timing critical designs. So its upto designers to decide which cells to use/avoid. I leave it to experts approval to merge or skip. Thanks

I don't understand. Unless the resizer has bugs, then excluding size 1 cells will do nothing for timing.

PR #316, opened 2022-11-28
update opnelane gf180 base config.tcl

kareefardi:
\+ set TECH_LEF_MIN and TECH_LEF_MAX for multicorner spef extraction

### DIFF
--- a/gf180mcu/openlane/config.tcl
+++ b/gf180mcu/openlane/config.tcl
@@ -22,7 +22,9 @@ set ::env(LIB_SLOWEST) "$::env(PDK_ROOT)/$::env(PDK)/libs.ref/$::env(STD_CELL_LI
 set ::env(LIB_TYPICAL) $::env(LIB_SYNTH)
 
 # Technology LEF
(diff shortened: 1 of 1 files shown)


RTimothyEdwards (approved):
Okay, but I think that RT_MIN_LAYER should be Metal1.  The fact that the standard cells use a lot of metal1 is irrelevant.  By forcing the router up to metal2, you are prohibiting the router from making simple point-to-point connections between neighboring cells, and therefore _increasing_ congestion, not _decreasing_ it.

kareefardi:
> Okay, but I think that RT_MIN_LAYER should be Metal1. The fact that the standard cells use a lot of metal1 is irrelevant. By forcing the router up to metal2, you are prohibiting the router from making simple point-to-point connections between neighboring cells, and therefore _increasing_ congestion, not _decreasing_ it.

I don't disagree. But a quick experiment yielded better results with setting min RT_MIN_LAYER to Metal2. It needs more investigation.

RTimothyEdwards:
Pulled and merged.  Not mirrored to github yet.

PR #317, opened 2022-12-01
bump magic version for gf180mcu

kareefardi:
I am getting `Error:  Magic version 8.3.348 is required by this techfile, but this version of magic is 8.3.346.` with the current version set in the json file.


RTimothyEdwards (approved):
Ah, yes, I last updated the JSON just before the update to magic.  Thanks.

RTimothyEdwards:
Pulled and merged.

PR #323, opened 2022-12-05
update gf180 9t openlane config:

kareefardi:
set the right site
set padding similar to 7t


RTimothyEdwards:
Pulled and merged.  Not mirrored yet.

RTimothyEdwards:
This was merged several days ago, so closing the ticket.

PR #328, opened 2022-12-14
sky130_fd_sc_hd Missing Pin Label Patches

stdavids:
This PR fixes the sky130_fd_sc_hd GDS which has some pin shapes that are missing labels which can cause issues with RC extraction. It uses klayout  to merge a patch GDS file with the main sky130_fd_sc_hd.gds file. The patch GDS is a minimal GDS file which just contains the pin label objects.


RTimothyEdwards:
I have opted for another approach, which is to ignore pin layer geometry when there is no text to go along with it.  Especially since text labels aren't placed at every single possible P&R grid point on all cells, there doesn't seem to be any particular point in the extra labels, and they just force the tools to do extra processing to accommodate them.  I apologize for dismissing your contribution, though, because it is well-thought out and well written.

stdavids:
Thank Tim and no worries, I appreciate your effort helping resolve this issue regardless of the approach!

PR #338, opened 2023-01-06
Ci fixes

antonblanchard:
This gets the CI closer to working again. One remaining problem is in the miniconda installer https://github.com/conda/conda/issues/10431


mithro:
I assume the issue is the SkyWater PDK? If we update the `third_party/make-env` submodule to the latest version that should work again -> https://github.com/SymbiFlow/make-env/pull/31

mithro:
https://github.com/google/skywater-pdk/pull/408 should fix the problem?

antonblanchard:
Thanks @mithro it does, and the CI is now passing with the skywater make-env fix and this PR. Thanks!

RTimothyEdwards (approved):
I don't really understand how this affects the CI operation, but I'll take your word for it.

RTimothyEdwards:
Pulled and merged (but not yet mirrored to github).

PR #347, opened 2023-02-02
Klayout XOR variables

kareefardi:
add KLAYOUT_XOR_IGNORE_LIST
add variable for klayout lmp file

Need confirmation that it is not an issue to ignore `81/14` during xoring gds files from magic and klayout (done as verification step in openlane). As far as I understood, this layer is added by magic and only used by magic for tap distance check ?


RTimothyEdwards (approved):
Looks good to me.

RTimothyEdwards:
Pulled and merged.  The github mirror will update overnight.

PR #349, opened 2023-02-07
openlane PDNsim changes

kareefardi:
\+ add `VDD_PIN_VOLTAGE` and `GND_PIN_VOLTAGE` needed by PDNsim for ir drop analysis
\- remove `sky130_fd_sc_hvl__tt_025C_3v30_lv1v80` from `LIB_SYNTH` as it breaks abc in yosys and I don't think level shifters are needed for synth (if I understand correctly that this lib is for level shifters)

### DIFF
--- a/gf180mcu/openlane/config.tcl
+++ b/gf180mcu/openlane/config.tcl
@@ -4,6 +4,9 @@ set ::env(DEF_UNITS_PER_MICRON) 2000
 
 set ::env(VDD_PIN) "VDD"
 set ::env(GND_PIN) "VSS"
+set ::env(VDD_PIN_VOLTAGE) "5.00"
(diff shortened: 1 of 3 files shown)


RTimothyEdwards (changes requested):
It appears that in `sky130_fd_sc_hvl/config.tcl` you are setting the pin voltage variables inside the EF_FORMAT block (which is used by nobody).  It should be moved outside the block.  Only things (mainly paths) that are different between EF_FORMAT = TRUE and EF_FORMAT = FALSE should be inside such an #ifdef block.

kareefardi:
Thanks for catching that. Update pushed

RTimothyEdwards (approved):
Looks good to me.

RTimothyEdwards:
Pulled and merged.  Mirror site will update tonight.

PR #365, opened 2023-04-16
lib: Make *_nointpwr.lib well-formed files

dlmiles:


RTimothyEdwards (approved):
Clearly needed, thanks.  Not sure where I got the versions that are in the code now but probably some attempt to parse the original sources lost the beginning of each comment.

RTimothyEdwards:
Pulled and merged.

PR #373, opened 2023-05-10
Fix sources clone

kareefardi:
- Add `; \` to chain commands to be able to reuse the shell variable created earlier. Each line of a make recipe is a separate shell instance.
- Escape `${FD_PV_COMMIT}`, and similar variables, to use the shell variable created in the recipe
- In `download.sh`, if git clone --branch fails, attempt to git clone and checkout - for commits.

### DIFF
--- a/scripts/download.sh
+++ b/scripts/download.sh
@@ -76,7 +76,8 @@ else
 	    else
 		# git clone $1 $2
 		# git checkout $3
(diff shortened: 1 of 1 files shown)


RTimothyEdwards (approved):
Ah, so there was a lot more than just the space around "=".

RTimothyEdwards:
Pulled and merged.

PR #377, opened 2023-06-02
gf180mcu: fix klayout lvs path

proppy:
With https://github.com/efabless/globalfoundries-pdk-libs-gf180mcu_fd_pr/commit/246beab385cfcc17e3b69bf7854e62c1ca79427c the `lvs` was removed from the `pr` repo after moving in `pv`, updating the makerfile to refer to the new location.

Related https://github.com/hdl/conda-eda/issues/333

PR #378, opened 2023-06-07
Update sources for gf180mcu

kareefardi:
https://github.com/RTimothyEdwards/open_pdks/pull/377 will fail the pdk build when using `--with-reference`. This PR updates the sources commit id to follow the changes done in the previously mentioned PR


RTimothyEdwards:
Looks good to me.

RTimothyEdwards:
Pulled and merged.

Issue #379, opened 2023-06-12
efabless io modules doesn't have a guard for USE_POWER_PINS


M0stafaRady:
some of `_ef_` modules under` sky130_fd_io.v` don't consider the case of simulating without `USE_POWER_PINS`.

RTimothyEdwards:
Shouldn't be too hard to do.  Create a pull request.

PR #380, opened 2023-06-12
Fix ef cell power guard

M0stafaRady:
fixes for 
https://github.com/RTimothyEdwards/open_pdks/issues/379


RTimothyEdwards (changes requested):
If you are moving the pins that are always defined to the end of the list, then please be consistent and do the same for the rewrite of `sky130_fd_io__top_gpiov2` (move all pins in the `ifdef` to the front of the list).  Also, please be consistent with indentation.  If your programming environment is not showing the difference in indentation between, say, lines 106 and 107, then please change your environment settings so that it does, or switch to a standard text editor like vim.

RTimothyEdwards (approved):
Looks good to me.

RTimothyEdwards:
Pulled and merged.

Issue #381, opened 2023-06-15
Soft connection checking uses a modified tech file that must be updated when the base file changes.


d-m-bailey:
@RTimothyEdwards We've talked about implementing the no-well extraction technology as a variant in the magic tech file.

Here are the base version and the no well version of the tech files I'm using for sky130A.

Is this enough information to create a variant? (maybe `nowell`)

[tech.tgz](https://github.com/RTimothyEdwards/open_pdks/files/11755146/tech.tgz)

RTimothyEdwards:
The device terminals are a general problem in that they are always "soft" connections---there is always some distance between the point under the device and the nearest well/substrate tap.  This is generally enforced by DRC rules, although the DRC rules tend to be rather computationally expensive.  I think there should be a way for the connectivity function to create for itself a different connectivity table which excludes the well and substrate types, and be able to apply that table selectively.  That could be set based on a criterion for type resistivity, to exclude the high-resistance well and substrate types automatically.  I have to look at the definition of the resistance calculation (the one in "extract", not "extresist") to see if it is sufficient to make a "soft" threshold based on wire resistance, which could be used to find soft connections through long poly or local interconnect wires.

RTimothyEdwards:
@d-m-bailey :  This would be a sort of coarse-grained check, but I'd like to know for layouts that have soft connection errors and those that use poly or local interconnect as long resistive routes, whether they show up clearly in the `.ext` file in the `node` statements for the net in question.  The `node` line syntax starts out `node <name> <R> <C> ...` where `<R>` is a resistance calculated from the area and perimeter of each layer in the net.  It should catch soft errors, but I just don't know how reliably.  The calculation of `<R>` assumes a simple point-to-point wire, and anything with branches or loops or other complex geometry is going to have an inaccurate result.

d-m-bailey:
@RTimothyEdwards I think I have a solution using variants. I'll post it soon.

d-m-bailey:
Should be fixed by #382

PR #382, opened 2023-06-17
Changes for LVS and softcheck 

d-m-bailey:
**magic technology**
Added nowell variants to magic technology to extract with no well/substrate connections.

**netgen setup**
Removed check for GDS extraction to ignore empty cells.
Made parallelization for fill/tap/decap cells more generic. 
Ignore prefixes (`XX_`) and suffixes (`$0`) when comparing layout to source.

### DIFF
--- a/gf180mcu/magic/gf180mcu.tech
+++ b/gf180mcu/magic/gf180mcu.tech
@@ -1582,7 +1582,7 @@ cifinput
 # NOTE:  All values in this section MUST be multiples of 25 
 # or else magic will scale below the allowed layout grid size
 
-style  import
+style  import variants (),(nowell)
  scalefactor 50 nanometers
  gridlimit 5
 
@@ -1594,7 +1594,14 @@ style  import
  ignore CAPDEF
  ignore VTEXT
  ignore FILLOBS
+variants (nowell)
+ ignore DNWELL
+ ignore SUBCUT
+ ignore NWELL
+ ignore PWELLTXT
+ ignore NWELLTXT
 
+variants ()
  layer pwell PWELL,PWELLTXT
  labels PWELL
  labels PWELLTXT port
@@ -1609,6 +1616,7 @@ style  import
  layer isosub SUBCUT
  labels SUBCUT
 
+variants *
  templayer ndiffarea DIFF
  and-not POLY
  and-not NWELL
@@ -3488,7 +3496,7 @@ end
 #-----------------------------------------------------
 
 extract
- style ngspice variants (),(hrhc),(lrhc),(hrlc),(lrlc)
+ style ngspice variants (),(hrhc),(lrhc),(hrlc),(lrlc),(nowell)
  cscale 1
  lambda  5.0
  units	microns
@@ -3597,7 +3605,7 @@ extract
 
 # Layer resistance
 
-variants ()
+variants (),(nowell)
 
 # Resistances are in milliohms per square
 # Optional 3rd argument is the corner adjustment fraction
@@ -4683,6 +4691,7 @@ variants *
 #
 # **The top metal may be any of metal3 to metaltp, depending on the stackup
 
+variants ()
  device msubcircuit pfet_03v3 pfet pdiff,pdc pdiff,pdc allnwell error \
 	l=l w=w a1=as p1=ps a2=ad p2=pd
  device msubcircuit nfet_03v3 nfet ndiff,ndc ndiff,ndc allpsub error \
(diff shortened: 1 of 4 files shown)


d-m-bailey:
Should resolve #381

d-m-bailey:
@RTimothyEdwards The changes suggested seem to cause the `nowell` variant to extract parameters 10000 times smaller.
```
  CIF input style is now "sky130()"                                                    |  Input style sky130(nowell): scaleFactor=2, multiplier=2                              
  -------------------------------------------------------------------------------------|  CIF input style is now "sky130(nowell)"                                              
```
Here's the ext file diff
```
  style ngspice()                                                                      |  style ngspice(nowell)                                                                
  scale 1000 1 500000                                                                  |  scale 1000 1 50                                                                      
 ```
And the property mismatch from `lvs.report`.
```Netlists match uniquely with property errors.
sky130_fd_pr__nfet_01v8:1 vs. sky130_fd_pr__nfet_01v8:1:
 l circuit1: 4.73   circuit2: 0.000473   (delta=200%, cutoff=1%)
 w circuit1: 0.55   circuit2: 5.5e-05   (delta=200%, cutoff=1%)
sky130_fd_pr__pfet_01v8_hvt:0 vs. sky130_fd_pr__pfet_01v8_hvt:0:
 l circuit1: 4.73   circuit2: 0.000473   (delta=200%, cutoff=1%)
 w circuit1: 0.87   circuit2: 8.7e-05   (delta=200%, cutoff=1%)
```
From the tech file, it looks to me that the scale factors should be the same for `()` and `(nowell)`.
```
-----------------------------------------------------------------------
cifinput
#-----------------------------------------------------------------------
# NOTE:  All values in this section MUST be multiples of 25
# or else magic will scale below the allowed layout grid size
#-----------------------------------------------------------------------

style  sky130 variants (),(vendor),(nowell)
 scalefactor 10 nanometers
 gridlimit 5
```
and the `extract` section
```
extract
 style ngspice variants (),(orig),(si),(hrhc),(lrhc),(hrlc),(lrlc),(nowell)
 cscale 1
 # NOTE: SkyWater SPICE libraries use .option scale 1E6 so all
 # dimensions must be in units of microns in the extract file.
 # Use extract style "ngspice(si)" to override this and produce
 # a file with SI units for length/area.

 variants (),(orig),(hrhc),(lrhc),(hrlc),(lrlc),(nowell)
 lambda  1E6
 variants (si)
 lambda 1.0
 variants *
```

d-m-bailey:
@RTimothyEdwards Please ignore the previous comment. There was an omission in the `sky130B.tech` file I was using that doesn't match then quoted `sky130A` version quoted above.

RTimothyEdwards:
@d-m-bailey :   In sky130, you removed the entire section of the netgen setup file that ignores the fill and tap cells.  But isn't that section necessary because those cells get removed from the extracted layout and therefore can never compare correctly to a verilog gate level netlist that has them?

Then in the gf180mcu setup, the fill and tap cells are always ignored, which is followed by an attempt to parallel combine them, which should never work because they've been ignored.

I'm not sure what the best solution is, but it needs to work for several different ways of doing extraction.  The original reason for the "ifdef" was that when extracting from a layout in .mag format using the PDK libraries, the fill and tap cells could be retained by marking them as abstract (or all cells can be marked as abstract, by using the abstract views).  Since the marking is not a GDS layer, then it is not preserved when writing GDS, so when reading back layout from GDS, then cells are not abstract (cell replacement could be done, but that would be making broad assumptions about the data in the GDS being the same as the data in the PDK, which is generally unwarranted), and cells like fill and tap will vanish upon being extracted.  So I don't see any way around needing to treat these two extraction cases differently in the netgen setup.

d-m-bailey:
@RTimothyEdwards Thanks for checking. The LVS setup that we're adding to precheck has the following parameter (from [lvs_config.md](https://github.com/d-m-bailey/extra_be_checks/blob/oeb-check/lvs_config.md))

- EXTRACT_ABSTRACT : List of cells to extract as abstract devices. Normally, cells that do not contain any devices will be flattened during netlisting. Using this variable can prevent unwanted flattening of empty cells. This has no effect of cells that are flattened because of a small number of layers. Internal connectivity is maintained (at least at the top level).

The base `lvs_config.json` sky130 file is 
```
$ head -50 tech/sky130A/lvs_config.base.json
{
        "EXTRACT_FLATGLOB": [
                "*sky130_fd_pr__*[A-Z]*"
        ],
        "EXTRACT_ABSTRACT": [
                "*__fill_*",
                "*__fakediode_*",
                "*__tapvpwrvgnd_*"
        ],
        "LVS_FLATTEN": [
                ""
        ],
        "LVS_NOFLATTEN": [
                ""
        ],
        "LVS_IGNORE": [
                ""
        ],
        "LVS_SPICE_FILES": [
                "$PDK_ROOT/$PDK/libs.ref/$STD_CELL_LIBRARY/spice/*.spice"
        ],
        "LVS_VERILOG_FILES": [
                ""
        ]
}
```
This causes the specified cells to be extracted as black-boxes and allows them to be compared during LVS.

The `MAGIC_EXT_USE_GDS` variable in the current setup file was something that I added to openlane when adding the `-lvs` option to `flow.tcl`. We're moving away from that largely unknown, undocumented and inflexible option to a more robust version in precheck. That is why I removed it.

As far as the `gf180mcu` fill devices being ignored, that was an oversight on my part. Since the `endcap` and `filltie` cells exist in the gate level verilog and are extractable as abstract devices, it is not necessary to ignore them. I'll test this and make another PR later.

The `fill` cells also exist in the gate level verilog, but get flattened during extraction due to the `gds flatten yes` option that flattens cells with few layers. Is there a way to exclude specific cells from the gds flattening process?

PR #391, opened 2023-08-08
Fix configure silent error

kareefardi:
Change realpath in configure.ac to handle empty argument and regenerate configure. Bug triggered by --disable-<>

### DIFF
--- a/scripts/configure
+++ b/scripts/configure
@@ -638,9 +638,9 @@ pkgpyexecdir
 pyexecdir
 pkgpythondir
 pythondir
-PYTHON_PLATFORM
 PYTHON_EXEC_PREFIX
 PYTHON_PREFIX
+PYTHON_PLATFORM
 PYTHON_VERSION
 PYTHON
 target_alias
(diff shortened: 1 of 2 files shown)


RTimothyEdwards:
But I tested with "--disable-" and it works fine.

kareefardi:
I am not sure why. On my machine this happens:
```
❯ cat configure
realpath() {
    python3 -c 'import os,sys;print(os.path.realpath(os.path.expanduser(sys.argv[1])))' $1
}

var=""
realpath $var
❯ bash configure
Traceback (most recent call last):
  File "<string>", line 1, in <module>
IndexError: list index out of range
❯ bash --version
GNU bash, version 5.1.16(1)-release (x86_64-pc-linux-gnu)
Copyright (C) 2020 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
```
Similar code section is executed in `configure` when `--disable-` is supplied.

RTimothyEdwards (approved):
Approved, although "configure" is a derived file, so I'm going to implement the 1-line change in "configure.ac" manually and update instead of merging the pull request.

PR #400, opened 2023-09-14
Fix Power Pins of `sky130_ef_sc_hd`

### LINKED ISSUE #397: Efabless cell LEF files do not correctly classify `VNB`, `VPB` as power pins
Consider this part of the LEF for `sky130_ef_sc_hd__decap_12`:

https://github.com/RTimothyEdwards/open_pdks/blob/ebba698e38217b7af5817de5ff6e7b1f4be45fa1/sky130/custom/sky130_fd_sc_hd/lef/sky130_ef_sc_hd__decap_12.lef#L24-L36

Designs from the PDK proper lack those pins entirely, but when `open_pdks` "injects" the pins, they're injected as follows:

```lef
PIN VNB
    DIRECTION INOUT ;
    USE GROUND ;
    PORT
      LAYER pwell ;
        RECT 0.005 0.785 0.925 1.015 ;
        RECT 0.005 0.105 3.590 0.785 ;
        RECT 0.150 -0.085 0.320 0.105 ;
    END
  END VNB
  PIN VPB
    DIRECTION INOUT ;
    USE POWER ;
    PORT
      LAYER nwell ;
        RECT -0.190 1.305 3.870 2.910 ;
    END
  END VPB
```

`USE POWER ;` and `USE GROUND ;` are used by OpenROAD to determine if the pins in question are power pins or not, so when writing an unpowered netlist, virtually every `sky130_ef_sc_hd__decap_12` is broken.

This looks like a straightforward fix. Should I submit a PR or is there a reason behind this?

donn:
Adds correct direction and power attributes to LEF views of `sky130_ef_sc_hd` cells.

I also added a Readme with your Slack comment for documentation purposes.

---

Resolves #397

### DIFF
--- a/sky130/custom/sky130_fd_sc_hd/Readme.txt
+++ b/sky130/custom/sky130_fd_sc_hd/Readme.txt
@@ -1,0 +1,3 @@
+These decap cells were redesigned to cut back on its use of the LI layer because the corresponding SkyWater cell was covered in something like 70-80% LI, and using enough of the decap in any one design would cause an LI density error.
+
+The fill cells were redesigned to add tap diffusion in the middle because otherwise the spacing of the nwells and the height of the cells prevents FOM fill, resulting in a different density error.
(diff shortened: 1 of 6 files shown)


RTimothyEdwards (approved):
Looks good to me!

RTimothyEdwards:
Pulled and merged.

donn:
@RTimothyEdwards Eh. It's not hurting anybody and we've already instructed OpenLane to not use it, so.

Issue #413, opened 2023-11-01
Weird cell LEF definitions for sky130


donn:
1. All EF cells
  * Lack a `SITE` definition
  * Are improperly(?) marked `CLASS BLOCK` instead of `CLASS CORE` as normal standard cells are

```lef
MACRO sky130_ef_sc_hd__decap_12
  CLASS BLOCK ;
  FOREIGN sky130_ef_sc_hd__decap_12 ;
  ORIGIN 0.000 0.000 ;
  SIZE 5.520 BY 2.720 ;
  PIN VGND
  …
```

2. All double-height cells in at least `fd_sc_hd`
  * Have their sites declared as `unithd` instead of the correct `unithddbl`, e.g.

```lef
MACRO sky130_fd_sc_hd__lpflow_lsbuf_lh_hl_isowell_tap_1
  CLASS CORE WELLTAP ;
  FOREIGN sky130_fd_sc_hd__lpflow_lsbuf_lh_hl_isowell_tap_1 ;
  ORIGIN 0.000 0.000 ;
  SIZE 6.440 BY 5.440 ;
  SYMMETRY X Y R90 ;
  SITE unithd ;
  …
```  

> When manually instantiated, those double-height cells will fail with the message "taller than any row" in OpenROAD for example, unless OpenROAD is instructed to explicitly construct double-height rows to accommodate them.

RTimothyEdwards:
For the cells in (1), fix them the way you like in `sky130/custom/sky130_fd_sc_hd/lef` and send me a pull request.
For the cells in (2), fix them in the https://github.com/efabless/skywater-pdk-libs-sky130_fd_sc_hd repository and send a pull request.

donn:
https://github.com/efabless/skywater-pdk-libs-sky130_fd_sc_hd/pull/12
https://github.com/RTimothyEdwards/open_pdks/pull/414

PR #414, opened 2023-11-02
Fix `LEF` views of EF cells

donn:
* Correctly set the `CLASS` property to `CLASS CORE SPACER` instead of `CLASS BLOCK`
* Set missing `SYMMETRY` property to `X Y R90` to match `fd_sc_hd` fill cells
* Set missing `SITE` property to `unithd`
* Created missing `sky130_ef_sc_hd__fill_4.v`
* Remove `fakediode`: Maintaining it is no longer worth it

---
Resolves (1) of https://github.com/RTimothyEdwards/open_pdks/issues/413


RTimothyEdwards (approved):
All of this looks good to me.

RTimothyEdwards:
Pulled and merged to opencircuitdesign.com.

PR #418, opened 2023-11-22
Sky130A OpenLane KLayout DRC support

kareefardi:
~This is a Draft PR. Don't merge yet.~
Edit:
This is good for review and merge now.

Edit 2:

\+ Add KLAYOUT_DRC_OPTIONS for integration with OpenLane
\~ Rename `KLAYOUT_DRC_TECH_SCRIPT` to `KLAYOUT_DRC_RUNSET` to be consistent with terminology found in KLayout documentation.


kareefardi:
@marwaneltoukhy @donn Please review

marwaneltoukhy:
LGTM

RTimothyEdwards:
@kareefardi @marwaneltoukhy @donn :  How does this change make any sense?  It's not just renaming KLAYOUT_DRC_TECH_SCRIPT to KLAYOUT_DRC_RUNSET, but it's doing it _only_ for sky130A, and leaving it undefined for sky130B.

kareefardi:
@RTimothyEdwards as far as I know we don't have a KLayout DRC deck for sky130B.
And in OpenLane we can make optional PDK variables that may or may not be defined.

RTimothyEdwards:
@kareefardi :  If you ran the existing klayout DRC deck on a sky130B design, then you would simply miss rules regarding ReRAM but catch everything else.  So instead, you would rather not run it at all and fail to provide any DRC results back to the user, instead of providing them with 99.5% of the checks?

kareefardi:
@RTimothyEdwards Ok make sense. Thank you. I removed the if condition allowing only sky130A

RTimothyEdwards (approved):
Okay, looks good now.

RTimothyEdwards:
Pulled and merged on opencircuitdesign.com.

PR #424, opened 2024-01-03
Change RT_CLOCK_MIN_LAYER to met3 for sky130

kareefardi:
This instructs the global router to use met3 and above for generating guides for the clock tree.


RTimothyEdwards (approved):
Looks okay to me.

PR #425, opened 2024-01-08
Change Maximum Capacitance Value

donn:
2 pF -> 0.2 pF

Per @passant5, this is a more realistic value


RTimothyEdwards (approved):
Looks good to me.

PR #426, opened 2024-01-10
Fix OpenLane Config for `gf180mcu`

donn:
Version 1.0.462 changed the directory name for lib files from `liberty` to `lib`, but this was not reflected in the OpenLane config, which causes the liberty files to not be found.

I've also taken the liberty (pun not intended) of updating the path in `qflow/gf180mcu.sh`, just in case.


RTimothyEdwards (approved):
Looks good to me.

RTimothyEdwards:
Pulled and merged.

RTimothyEdwards:
And mirrored.

PR #441, opened 2024-05-08
Fix source cloning for sky130

donn:
#373, specifically commit `db529a71335176db64b18b98228fdd7f03a8c24b` fixed an issue where the shell commands inside `gf180mcu/Makefile.in` were not properly chained for some make productions, causing the branch commit to be dropped, however, the same issue persisted for `sky130`. This caused builds to be non-reproducible for `sky130` where the Makefile would *always* grab the latest for each repo. The result was `--with-reference` never worked for sky130.

This also adds an additional print to `scripts/download.sh` just to verify the commit is correctly propagated to the `git clone` commands.

### DIFF
--- a/scripts/download.sh
+++ b/scripts/download.sh
@@ -80,7 +80,11 @@ if [ "${1: -3}" == ".gz" ] ; then
 else
 
     if type "git" > /dev/null; then
-        echo "Cloning $1 to $2"
+        branch_commit_msg=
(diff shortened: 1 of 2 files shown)


RTimothyEdwards (approved):
All looks good, thanks.

RTimothyEdwards:
Pulled and merged.

PR #454, opened 2024-09-16
sky130: Add a new sky130_ef_sc_hd__conb_1 to fix npc.2 DRC errors

smunaut:
The current conb cell has an NPC layer that's not entirely compatible with other cells in the library and can cause DRC failure depending on what's put left / right of it.

Draft PR as this currently doesn't include any '.lib' files that are needed so flow ends with 
```
[IFP-0039] Liberty cell or port sky130_ef_sc_hd__conb_1/LO not found.
```


RTimothyEdwards:
Does it make more sense to just fix the `sky130_fd_sc_hd` version of the cell (which avoids the .lib file issue)?
Has any check been done to make sure that the fixed version correctly abuts all other cells in all positions/orientations?

smunaut:
Fixing `sky130_fd_sc_hd` definitely can make sense, I was just under the impression that the `_fd_` cells had to remain exactly what we got from skywater.

I did check that the fixed version should have a `npc` layer compatible with *most* the other cells in the library ( Wrote some `gdspy` script that was checking that ). The only cell I found to be incompatible is `and2_0` . It's incompatible both before and after the changes, that cell also has weird `npc` geometry which makes it incompatible with a lot of other cells.

I also patched the cell in a few designs we had in tiny tapeout that were exibiting the issue and that fixed it without causing any new DRC errors.

I didn't check any other layers ( `nsdm` / `psdm` / ... ) because I didn't touch those and we didn't have any reports of errors with those.

RTimothyEdwards:
@smunaut :  I do not like to make changes to the sky130 PDK unnecessarily, but this is a long-standing error that I am well aware of and would like to have a patch for in the upstream repo.  A fix to `and2_0` would be helpful as well.

RTimothyEdwards:
A fix as a PR to https://github.com/efabless/skywater-pdk-libs-sky130_fd_sc_hd is preferred.

smunaut:
Ack, I'll see what I can do for `and2_0` and open a PR against the above repo instead.

PR #461, opened 2024-10-16
Make copied source read-only files writable (fixes #460)

### LINKED ISSUE #460: Build fails when sources are in a read-only directory
I'm using NixOS where packages are built in isolated sandboxes and the build artifacts are stored in read-only locations. For this reason, and to be able to select libraries in various combinations, I build a package per library. When `open_pdk` copies the sources to the staging area the files are still read-only and the following scripts fail to open some files for writing:
- rename_models.py
- rename_cells.py
- inc_verilog.py
- fix_io_lef.py
- inc_verilog.py
- foundry_instal.py

All of the above can be fixed by including
```python
import stat

def makeuserwritable(filepath):
    if os.path.exists(filepath):
        st = os.stat(filepath)
        os.chmod(filepath, st.st_mode | stat.S_IWUSR)
```
and before opening for write
```diff
         try:
+            makeuserwritable(outname)
             with open(outname, 'w') as outFile:
                 for i in fixedlines:
                     print(i, file=outFile)
```

The same problem affects `sky130/Makefile.in` and can be corrected by
```diff
 	# then remove the larger of the two example directories 
 	if test "x${XSCHEM_PATH}" != "x" ; then \
 		cp -rp ${XSCHEM_PATH}/* ${XSCHEM_STAGING_$*} ; \
+		chmod -R u+w ${XSCHEM_STAGING_$*}; \
 		rm -rf ${XSCHEM_STAGING_$*}/decred_hash_macro ; \
 	fi
 	# xschem setup is for sky130A.  Fix for the given target variant.
```

The full patches can be inspected in [fbe-flakes](https://github.com/fedeinthemix/fbe-flakes/tree/main/pkgs/development/pdk/open-pdks). If there is interest I can prepare a PR.

fedeinthemix:
As discussed in #460

### DIFF
--- a/common/foundry_install.py
+++ b/common/foundry_install.py
@@ -401,6 +401,7 @@ def vfilefilter(vfile):
         # File contents have been modified, so if this file was a symbolic
         # link, then remove it.  Otherwise, overwrite the file with the
         # modified contents.
+        makeuserwritable(vfile)
         if os.path.islink(vfile):
             os.unlink(vfile)
         with open(vfile, 'w') as ofile:
--- a/gf180mcu/custom/scripts/inc_verilog.py
+++ b/gf180mcu/custom/scripts/inc_verilog.py
@@ -13,6 +13,12 @@
 import re
 import os
 import sys
+import stat
+
+def makeuserwritable(filepath):
+    if os.path.exists(filepath):
+        st = os.stat(filepath)
+        os.chmod(filepath, st.st_mode | stat.S_IWUSR)
 
 def filter(inname, outname):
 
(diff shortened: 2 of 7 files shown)


RTimothyEdwards (approved):
I have to wonder why the source files must end up as read-only, but at least there is no harm done in making sure all copied files are writeable.

RTimothyEdwards:
Pulled and merged in opencircuitdesign.com (github mirror will update in about 20 hours).

RTimothyEdwards:
@fedeinthemix :  But doesn't that mean there are numerous other places where `chmod` needs to be applied in the Makefile?   You changed one line for xschem sources in one of the PDKs, but there is a lot more copying going on than that.

fedeinthemix:
@RTimothyEdwards  As long as the files are not modified it's OK. At the moment the changes I proposed are enough to make the sky130 PDK installation complete.

It's quite possible that more changes will be needed for other PDKs, but so far I've not packaged them up.

PR #468, opened 2024-11-27
Remove entry to XSCHEM_LIBRARY_PATH

mole99:
Since the current directory of the xschemrc is now appended to XSCHEM_LIBRARY_PATH via `[file dirname [info script]]`, it is no longer necessary to add `${PDK_ROOT}/${PDK}/libs.tech/xschem`.

This prevents a double entry of the PDKs directory when inserting a symbol in xschem.


RTimothyEdwards:
This has been merged, although I don't recall exactly when.  Anyway, the PR can be closed.

PR #473, opened 2025-01-29
Replace CI

donn:
This removes the technical debt-laden CI with a new one based on Nix and Volare.

The CI uses Magic 8.3.515 and Volare 0.20.6 to build both sky130 and gf180mcu. A successful build is required for the CI to clear. This will run on all pushes and pull requests, and it takes about 30 minutes.

In the event a new tag is pushed and the secret `BOT_GH_TOKEN` is specified, the result of the build is pushed to Volare as a pre-release. Pre-releases may be converted to full releases at the Volare maintainer's discretion.

---

Incidentally, the reason macOS was picked is that the Mac runners, presently, have way more available storage than the Linux runners, especially if you delete the 8 billion versions of Xcode installed.


RTimothyEdwards (commented):
Looks good to me.

RTimothyEdwards:
Noted that this PR has been closed and superceded by #474.  Will not proceed with a merge on this one.

PR #476, opened 2025-02-19
Fixed a drc error in sky130_fd_bs_flash__special_sonosfet_star

urish:
Increased gate_to_diffcont from 0.145 to 0.158, to fix the following DRC error: Diffusion contact to SONOS gate < 0.075um.

Tested and found 0.158 to be the minimal value that fixes this DRC issue.

### DIFF
--- a/sky130/magic/sky130.tcl
+++ b/sky130/magic/sky130.tcl
@@ -7145,6 +7145,7 @@ proc sky130::sky130_fd_bs_flash__special_sonosfet_star_draw {parameters} {
 	    id_surround		1.355 \
 	    min_effl		0.185 \
 	    min_allc		0.26 \
+	    gate_to_diffcont	0.158 \
     ]
     set drawdict [dict merge $sky130::ruleset $newdict $parameters]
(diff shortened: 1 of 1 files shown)


RTimothyEdwards:
Sorry I didn't notice this before today;  somehow it didn't come to my attention in February.
I will check the diffusion contact distance.  Technically, it cannot be 0.158 because 0.008um is not a multiple of the manufacturing grid of 5nm.  I assume this is because you were increasing the value by 1nm and 0.158 was the point where it rounded up.  The actual value should probably be 0.16.  But it's also possible that the DRC rule is wrong.

RTimothyEdwards:
Reviewing the documentation, 0.075um is _not_ rule `licon.11`, so I'm not sure if the rule distance was entered correctly or if there is a different rule for SONOS that should have been listed as the rule.

RTimothyEdwards:
I am going to close this, as SkyWater no longer allows users to use the SONOS layer at all since shortly after this PR was made.  Sad but true.

PR #486, opened 2025-08-11
Update workflow for ciel

mole99:
- Volare replaced by Ciel
- Use the Nix cache from the FOSSi Foundation
- Always use the latest version of the tools - if it fails we need to fix it anyways

Before merging, we need to update `BOT_GH_TOKEN` in the repo.

I'm still thinking about whether we should keep pushing new releases as pre-release or not? (This is still from Efabless.)
When do we decide that a new release is stable? Shouldn't open_pdks always contain the latest bug fixes and therefore be considered stable? Experimental features should probably happen in a development branch?

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -17,8 +17,8 @@ jobs:
               archs: "arm64",
             },
           ]
-        pdk: [sky130, gf180mcu]
-    name: Build ${{ matrix.pdk }} with Volare | ${{ matrix.os.name }} | ${{ matrix.os.archs }}
+        pdk-family: [sky130, gf180mcu]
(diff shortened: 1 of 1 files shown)


RTimothyEdwards:
I would not call anything pre-release.  To me, all releases are stable.  Virtually everything being done to the PDK is just minor bug fixes.  The last thing that might have been called "experimental" was the introduction of the new I/O cells into the sky130_fd_io library, and that was over a year ago.

RTimothyEdwards:
Pulled and merged.

mole99:
Too fast ^^ Or rather, it's my fault for not setting the PR to draft mode.

We still need to remove the pre-release argument from the script. I'll open another PR :)

PR #498, opened 2026-02-28
Add missing xpoly contact

leviathanch:
Skywater's high precision resistors require dedicated contacts and can't simply be contacted to with a simple poly contact.

While the DRC rules catch an attempt of contacting xpolyres with a normal poly contact, the contact definitions didn't reflect that.

### DIFF
--- a/sky130/magic/sky130.tech
+++ b/sky130/magic/sky130.tech
@@ -6860,6 +6860,7 @@ wiring
 #endif (METAL5)
 
  contact pc  170  poly 50 80 li 0 80
+ contact xpc 170 xpolyres 50 80 li 0 80
  contact pdc 170 pdiff 40 60 li 0 80
  contact ndc 170 ndiff 40 60 li 0 80
  contact psc 170   psd 40 60 li 0 80


RTimothyEdwards:
This is in the "wiring" section.  I've never considered using the "wiring" tool to draw resistors, and I don't think it would be a very good idea, anyway.  Much better to use the device generator to automatically create the resistor with the correct value and geometry.  I doubt that the wiring tool can really generate the slot contacts correctly, but I'm quite certain that the values you put there aren't going to work.

leviathanch:
Hi.
Using your TCL scripts isn't an option because I'm automatically generating
the geometries in Python and only at the end I output them into a magic file.
We're reading in yout .tech files, extract the design rules, create cells, 
place and route them and then write out a magic for doing a final DRC check
and a SPICE extraction

Cheers
/lev

RTimothyEdwards:
This seems like the worst possible way to use magic.  You are ignoring its strengths---Native scripting of layout routines---and using its weaknesses (DRC).  If you want to use Python to do layout, then use klayout or gdsfactory and write GDS out, use klayout to run full DRC, and then read the GDS into magic to do extraction.

Regardless, though, if you are not using magic to generate a .mag file, then why do you care what's in the wiring section of the tech file?

leviathanch:
That doesn't work because Magic doesn't properly import special device layers
from the GDS we feed it, so I decided to try it the other way around which
works properly and produces diodes, resistors, capacitor and transistors
which are properly recognized by Magic's SPICE extraction and the GDS export
also passes through klayout's DRC check.

We had to add that one line to our copy of the SKY130 tech file and Philip 
suggested I should try to upstream it.

Logically xpc fits in there perfectly and our structures all work out fine, all
it needed for it to work with SKY130 was that one line.

RTimothyEdwards:
But the wiring section has absolutely no impact on anything that you're doing, so you seem to be relying on sections of the tech file to tell you things about the process that should not be inferred.  And this seems like a complete ass-backwards hack simply to avoid learning how to use magic properly.

leviathanch:
You obviously don't understand what I'm doing here.

The LibrePDK device generator takes in a generic JSON netlist with
fine tuning parameters for the circuit specific to known technology nodes
then generic device generators generate the geometry for resistors,
capacitors, diodes and transistors, place and route it and then write
it out as a Magic file.

https://gitlab.libresilicon.com/generator-tools/librepdk

It's just easier to use Magic for managing the design rules because
folks seem to be providing a complete and easy to parse list of design rules.

https://gitlab.libresilicon.com/generator-tools/librepdk/-/blob/
7a22b56d0fd02ec6e7bb95ff41f69d79481d3a4e/LibrePDK/technologies/magic_tech/
SKY130/libresilicon.tech

For now Magic does the trick.

Philip just suggested I should "contribute back to the project", so that PR
was more a sign of good faith.

I guess as soon as Magic won't do it anymore, I won't bother upstreaming
more patches, and just hard fork the parts of the code in Magic I need and
integrate it into LibrePDK directly. I got the memo.

Upside is that since we're still using the same format I can just keep copying
new technologies you get over into my repo and use it there.

/lev

RTimothyEdwards:
Ah, I see.  You have a complete place and route flow.

I am just surprised that you think that magic's DRC rules are either complete or easy to parse---I would not have said either one, although I think on the whole they are probably a bit easier to parse than klayout's rules, but I don't really say that with any confidence.

I will be fair and take my usual approach here:  Since there doesn't seem to be any particular drawback to the inclusion of the resistor contact type in the wiring section, even though I think it's a bit strange to have it there, and since it supports your tools, I will go ahead and merge it.

RTimothyEdwards:
Pulled and merged to the opencircuitdesign.com repository;  the mirror copy to github will be done overnight tonight.

leviathanch:
Thanks for your understanding.
To explain a bit more why I'm going this route:
The whole point is precisely to "learn Magic VLSI", just that the AI has to learn it, not me.
In a few days we should be at a point where she can write out an entire hierarchical Magic VLSI file with routes and component includes.
After that she will evolve tcl scripts for doing what a human would do in Magic when tweaking the layout in order to optimize that properties when running the spice extraction in ngspice.
The goal is to run ```npn run dev``` and end up with an interface where you can ask the AI to design you a generic schematic for a 555 or so. Then you can select any of the supported available Magic VLSI technologies and the tools will generate a basic unoptimized layout for the chosen node. Then you can tell the system for what parameters it should optimize the layout and it runs through an evolutionary cycle where the Magic is tweaked, spice extracted, a simulation evaluated and matched against the fitness function. After that, finteuning parameters are stored for that circuit and specific node the evolution was successful.

PR #504, opened 2026-04-24
sky130/magic: Ensure minimum 1.27um width for RPM/URPM

smunaut:


RTimothyEdwards:
@smunaut :  Can you explain the purpose of this change?  (Other than I get that it is to ensure that RPM and URPM layers have a minimum width of 1.27um.)

You have changed only the temporary layer created for DRC checks, which means that it does not actually affect the RPM output in GDS, only the RPM temporarily generated for running DRC checks.  I don't think that's what you meant to do.

But I'm more worried that `grow-min` does strange things sometimes and I generally don't use it except for layers which I know will be only simple isolated rectangles.  My preferred solution is just to use mask-hints to ensure that the RPM and URPM layers are the right width.  I put the templayer `rpm_generate` in the `drc` style so that I could write a DRC rule that would flag when the layer is too narrow, leaving the user to fix it with a mask hint.

The generated device drawing script could (and probably should) detect when it has a narrow, single resistor that will not meet the RPM minimum width requirement.  It could also use this information to increase the distance between the resistor and the guard ring (if there is one) so that the RPM layer does not overlap any guard ring contacts.  That was an error on the last tapeout that caused a major headache;  in spite of there being no documented rule preventing RPM from overlapping diffusion or diffusion contacts, doing so causes SkyWater's own DRC deck (or maybe it was Cadence's?) to identify the diffusion contact as a poly contact instead and ends up raising multiple DRC errors.  I do not know if that error was ever fixed on the SkyWater (or Cadence) side, but it is probably for the best just to avoid the situation altogether.

RTimothyEdwards:
(Mask hints, by the way, _will_ be greatly improved in the near future as soon as I have some time to work on it.)

smunaut:
Ah yeah, I was trying to undo part of 9ca6f00b4360922e095033945f36198060b65086
Some of it was removed already by bb0ba76993fbea9cad2298a35d3c99723ad2bb62 but it was still causing issue.

I guess I didn't dig enough and this patch fixed my immediate issue/symptom but I misunderstood what it was doing.

Mainly :
* Loading a valid GDS ( with rpm/urpm wider than 1.27um ) still throws an error because magic regenerates it narrower. Since librelane runs magic DRC that means including a valid macro still throws a DRC error.
* A lot of the pcells for resistors generate invalid rpm/urpm layer causing users to be confused about what the problem is since (1) they used a pcell and expect it to produce valid output and (2) rpm/urpm is not even directly visible in magic and quite a few don't even know what those are.

smunaut:
See related issue I opened in librelane about using `gds maskhints yes` when reading GDS in for DRC : https://github.com/librelane/librelane/issues/928#issuecomment-4315508477

RTimothyEdwards:
It seems like the best solution is to rewrite some of the resistor device generator script to add the mask hint to ensure that the output is DRC clean.  Your solution (properly copied to the actual GDS generating output style) is the "quick fix", but I can probably get a proper fix to the resistor drawing routine in a day or so (given that I am making reasonable progress on my presentations for next week at Latch-Up).

smunaut:
Fixing the pcell won't fix the first problem with DRC though right ?

smunaut:
Closing since this doesn't seem to do what I thought it did ...

I still think magic should ensure rpm/urpm min-width but apparently this is not the way to do it :sweat:

PR #509, opened 2026-05-01
sky130: improve pin access in gpiov2_pad LEF

mole99:
This PR makes slight adjustments to the LEF of sky130_ef_io__gpiov2_pad, so that the pins are only located at the core side.

You can see the issue here:

Bildschirmfoto vom 2026-04-30 11-53-17

The physical pin for `IN` runs along the entire the side of the I/O pad. OpenROAD creates access points everywhere on the pin, which can lead to DRT getting confused during routing.

Therefore, in this PR, I capped the pin shapes of all signal pins to 0. This way, access points will only be created where it makes sense (near the core side). This seems to work much better in practice than having these long pins.

Note: I opened this PR so that I wouldn't forget about it, but it can wait until after Latch-Up.

### DIFF
--- a/sky130/custom/sky130_fd_io/lef/sky130_ef_io__gpiov2_pad.lef
+++ b/sky130/custom/sky130_fd_io/lef/sky130_ef_io__gpiov2_pad.lef
@@ -44,7 +44,7 @@ MACRO sky130_ef_io__gpiov2_pad
     USE SIGNAL ;
     PORT
       LAYER met3 ;
-        RECT 45.865 -2.035 46.195 34.770 ;
+        RECT 45.865 -2.035 46.195 0.000 ;
     END
   END ANALOG_POL
   PIN ANALOG_SEL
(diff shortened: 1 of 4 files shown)


RTimothyEdwards:
Did you make the change manually, or with specific options to magic's "lef write" command?  Because I have some scripts that create the views from .mag sources.  It is likely that I will not use those scripts again, but if there's a setting I can change to keep the output consistent, then at least I can do that.

mole99:
I did make the changes manually. It would definitely be better to extract the LEF with `-pinonly`, but it seems that some of the pads have pin shapes that do not fully match the metal shapes. For example, the pin shape for the bondpad pin is often just a small rectangle.

grafik

So maybe we leave `-pinonly` as a future improvement?

mole99:
I have now also added the sky130_ef_io__analog cells to the setup. This is because I needed sky130_ef_io__analog_esd_pad to control the ENABLE_H signals of the I/Os externally.

RTimothyEdwards:
I would recommend the `sky130_fd_io__top_analog_pad` for that.  It was hiding in older repository versions and wasn't available in the open PDK when I first made the custom analog pads, but since it was developed by SkyWater (or rather, Cypress), it is presumably better than mine.

RTimothyEdwards (approved):
Looks tood to me.

RTimothyEdwards:
Pulled and merged into the opencircuitdesign.com repository.

mole99:
Thanks!

Issue #511, opened 2026-05-04
latest sky180 builds do not work with latest Librelane


jeras:
I think this is related to changes to `config.json` variables `STA_CORNERS` and `DEFAULT_CORNER`.

LibreLane creates broken `config.json` files. See the difference between the last and previous `open_pdks` versions. The variables `DEFAULT_CORNER` and `LIB` are broken.

```
diff runs/latest-78f0ccbb45a0d560890540a1ebc5ccf4d8370c70/01-verilator-lint/config.json runs/latest-d815bb30c9afdf9e264c276a8a2b533108dea3d0/01-verilator-lint/config.json
6,8c6,8
<         "nom_*": "/home/izi/.ciel/ciel/sky130/versions/78f0ccbb45a0d560890540a1ebc5ccf4d8370c70/sky130A/libs.ref/sky130_fd_sc_hd/techlef/sky130_fd_sc_hd__nom.tlef",
<         "min_*": "/home/izi/.ciel/ciel/sky130/versions/78f0ccbb45a0d560890540a1ebc5ccf4d8370c70/sky130A/libs.ref/sky130_fd_sc_hd/techlef/sky130_fd_sc_hd__min.tlef",
<         "max_*": "/home/izi/.ciel/ciel/sky130/versions/78f0ccbb45a0d560890540a1ebc5ccf4d8370c70/sky130A/libs.ref/sky130_fd_sc_hd/techlef/sky130_fd_sc_hd__max.tlef"
---
>         "nom_*": "/home/izi/.ciel/ciel/sky130/versions/d815bb30c9afdf9e264c276a8a2b533108dea3d0/sky130A/libs.ref/sky130_fd_sc_hd/techlef/sky130_fd_sc_hd__nom.tlef",
>         "min_*": "/home/izi/.ciel/ciel/sky130/versions/d815bb30c9afdf9e264c276a8a2b533108dea3d0/sky130A/libs.ref/sky130_fd_sc_hd/techlef/sky130_fd_sc_hd__min.tlef",
>         "max_*": "/home/izi/.ciel/ciel/sky130/versions/d815bb30c9afdf9e264c276a8a2b533108dea3d0/sky130A/libs.ref/sky130_fd_sc_hd/techlef/sky130_fd_sc_hd__max.tlef"
12c12
<     "DEFAULT_CORNER": "nom_tt_025C_1v80",
---
>     "DEFAULT_CORNER": "nom_",
44,54c44
<     "LIB": {
[...]

PR #507, opened 2026-04-27
Update gf180mcu LibreLane configs and add eFuses

mole99:
This PR updates the gf180mcu LibreLane configs and adds the eFuses library.

- Add Egor's gf180mcu_re_efuse library
- Update LibreLane configs
  - Add latch maps (gf180mcu_fd_sc_mcu7t5v0/mcu9t5v0)
  - Add pad configs (gf180mcu_fd_io/gf180mcu_ocd_io)
  - Add scl config (gf180mcu_as_sc_mcu7t3v3)
- Add Verilog blackbox views for the I/Os and SRAMs
- Ensure padring generation is working properly
  - [x] Update gf180mcu_fd_io/gf180mcu_ocd_io hashes
- Use local LibreLane config for gf180mcu_as_sc_mcu7t3v3 for now
  - This is required because of the `SCL_LIBS` variable
  - Update the other variable names
- Add SYNTH_CLOCKGATE_* variables for clock gating
- Make use of the new KLayout DRC setup
  - [x] Update pr/pv hashes

PR #512, opened 2026-05-05
sky130: updated variables in CSL config.tcl files

jeras:
Update deprecated variables to conform to latest LibreLane:

* `PLACE_SITE_WIDTH` and `PLACE_SITE_HEIGHT` have been deprecated since OpenLane 2.0.0b16. Now automatically extracted from PLACE_SITE.
* `FP_WELLTAP_CELL` and `FP_ENDCAP_CELL` have been deprecated since OpenLane 2.0.0b17. The new variables have the `FP_` prefix removed.
* `SYNTH_DRIVING_CELL_PIN` has been deprecated, instead use `SYNTH_DRIVING_CELL` with format `{cell}/{port}`.
* added `SYNTH_CLK_DRIVING_CELL` variable set to `clkinv`.
* `SYNTH_MIN_BUF_PORT` was deprecated, instead use `SYNTH_BUFFER_CELL` with format `{cell}/{input_port}/{output_port}`.
* `SYNTH_TIEHI_PORT`/`SYNTH_TIELO_PORT` are deprecated, instead use `SYNTH_TIEHI_CELL`/`SYNTH_TIELO_CELL` with format `{cell}/{port}`.
* `CELL_CLK_PORT` has been deprecated since OpenLane 2.0.0a24, it is not used in the current version.
* `FILL_CELL` and `DECAP_CELL` were renamed to `FILL_CELLS` and `DECAP_CELLS` since LibreLane 3.0.0.
* `RE_BUFFER_CELL` is not used in LibreLane.
* Instead of `DIODE_CELL_PIN`, use `DIODE_CELL` with format `{cell}/{port}`.
* `FAKEDIODE_CELL` is not used by LibreLane.
* `ROOT_CLK_BUFFER` not used in LibreLane (probably supposed to do what `CTS_ROOT_BUFFER` does).
* `CTS_CLK_BUFFER_LIST` was deprecated, use `CTS_CLK_BUFFERS` instead.
* `FP_PDN_RAIL_WIDTH` was deprecated, use `PDN_RAIL_WIDTH` instead.
* `CTS_MAX_CAP` is unnecessary, LibreLane will extract it from `CTS_CLK_BUFFERS`.
* `CTS_CLK_BUFFER_LIST` was deprecated, instead use `CTS_CLK_BUFFERS`.

### DIFF
--- a/sky130/librelane/sky130_fd_sc_hd/config.tcl
+++ b/sky130/librelane/sky130_fd_sc_hd/config.tcl
@@ -10,49 +10,40 @@ set ::env(SYNTH_MUX_MAP) "$::env(PDK_ROOT)/$::env(PDK)/libs.tech/librelane/$::en
 # Placement site for core cells
 # This can be found in the technology lef
 set ::env(PLACE_SITE) "unithd"
-set ::env(PLACE_SITE_WIDTH) 0.460
-set ::env(PLACE_SITE_HEIGHT) 2.720
 
 # welltap and endcap cells
-set ::env(FP_WELLTAP_CELL) "sky130_fd_sc_hd__tapvpwrvgnd_1"
-set ::env(FP_ENDCAP_CELL) "sky130_fd_sc_hd__decap_3"
+set ::env(WELLTAP_CELL) "sky130_fd_sc_hd__tapvpwrvgnd_1"
+set ::env(ENDCAP_CELL) "sky130_fd_sc_hd__decap_3"
 
 # defaults (can be overridden by designs):
-set ::env(SYNTH_DRIVING_CELL) "sky130_fd_sc_hd__inv_2"
+set ::env(SYNTH_DRIVING_CELL) "sky130_fd_sc_hd__inv_2/Y"
 #capacitance : 0.017653;
-set ::env(SYNTH_DRIVING_CELL_PIN) "Y"
+#set ::env(SYNTH_CLK_DRIVING_CELL) "sky130_fd_sc_hd__clkinv_2/Y"
 # update these
 set ::env(OUTPUT_CAP_LOAD) "33.442" ; # femtofarad __inv_16 pin A cap (https://github.com/google/skywater-pdk-libs-sky130_fd_sc_hd/blob/main/cells/inv/sky130_fd_sc_hd__inv_16__tt_025C_1v80.lib.json)
-set ::env(SYNTH_MIN_BUF_PORT) "sky130_fd_sc_hd__buf_2 A X"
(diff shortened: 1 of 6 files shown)


jeras:
A few more comments:

0. I was unable to properly test this changes due to https://github.com/RTimothyEdwards/open_pdks/issues/511

1. this commit tried to avoid any functional changes, but is is possible there will be side effects due to:
   * LibreLane can be unpredictable when it comes to handling deprecated configuration variables, see https://github.com/librelane/librelane/issues/932
   * `CTS_MAX_CAP` was removed, and LibreLane is supposed to extract it from `CTS_CLK_BUFFERS` (which is a list and not a single cell) so the two values might differ

2. I have added `SYNTH_CLK_DRIVING_CELL` variables but I have commented them out to avoid functional changes.
   This adds a dedicated clock buffet to the general buffer `SYNTH_DRIVING_CELL`.

3. Cell lists `FILL_CELLS`, `DECAP_CELLS`, `CTS_CLK_BUFFERS` and `TRISTATE_CELLS` seem to be arbitrary, some contain an asterisk, some a limited subset without explanation. A special case would be `DIODE_CELL`, where LibreLane expects a single cell, but the SCL provides many options (I might ask LibreLane about this).

mole99:
@jeras Thank you, this looks like a thorough clean-up.

Regarding gf180mcu, I currently have a PR open that makes a number of changes: https://github.com/RTimothyEdwards/open_pdks/pull/507
I plan to finish it up soon.

Many of the cell lists still come from Efabless from around the OpenMPW time. Unfortunately, as there are no comments, the reasons for excluding or choosing specific cells are probably lost to time.
Some cells may be disabled due to DRC errors when abutting certain other cells.
Maybe @RTimothyEdwards still remembers the reasons for some of the cells?

RTimothyEdwards:
I remember the reasons for some of the cells, but on the whole, any of those reasons could have been long gone, and we need a thorough review of the contents of all the "exclude" lists, which are quite long and therefore presumably adversely affecting synthesis results.

mole99:
Yes, I think we need to bite the bullet and clear the exclude lists one final time. This time, we should make a note of the reasons why cells were excluded.

mole99 (approved):
LGTM

You've motivated me to cleanup the gf180mcu PDK in my other PR :)

RTimothyEdwards (approved):
The librelane configs aren't really my area of expertise, but I get the general gist of it and it looks okay to me.

RTimothyEdwards:
Pulled and merged to opencircuitdesign.com.  Github mirror will update overnight.

PR #514, opened 2026-05-17
Add support for postscript file generation

Ahmed-Nematallah:
Added support for postscript file generation be including a section based on 
http://opencircuitdesign.com/magic/tech.html
http://opencircuitdesign.com/magic/archive/scmos-sub.tech

Layers were automatically generated from the mos.24bit.dstyle and mos.24bit.std.cmap files along with the styles section in the current tech file. stipples and colors as well as layer order were preserved as best as possible.

### DIFF
--- a/gf180mcu/magic/gf180mcu.tech
+++ b/gf180mcu/magic/gf180mcu.tech
@@ -5362,5 +5362,162 @@ plot
      draw nwell cwell
      draw pwell cwell
      draw dnwell cwell
+
+  style postscript
+    # --- Stipple Pattern Matrices (16*16 pixels) ---
+     1 30003000 00000000 03C003C0 00000000 03C003C0 00000000 00000000 00000000
+     2 F0F0F0F0 F0F0F0F0 00000000 00000000 0F0F0F0F 0F0F0F0F 00000000 00000000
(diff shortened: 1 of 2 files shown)


RTimothyEdwards (approved):
Great work!  Automation is definitely the key here---I could never convince myself that it was worth the effort to add the PostScript plot configuration to the tech files.

RTimothyEdwards:
This PR has been pulled and merged into the repository at opencircuitdesign.com.  The github mirror will update automatically overnight.

RTimothyEdwards:
I have reverted this pull request because the tech file changes do not respect the #ifdef block definitions that determine how the tech file is written for each process variant.  Instead, it contains all layers including ones which may not exist in the specific variant being built, which causes magic to generate an error message on startup.

Ahmed-Nematallah:
> I have reverted this pull request because the tech file changes do not respect the #ifdef block definitions that determine how the tech file is written for each process variant. Instead, it contains all layers including ones which may not exist in the specific variant being built, which causes magic to generate an error message on startup.

@RTimothyEdwards I apologize for this, I will fix the conversion script to include the #ifdef block definitions and upload the script and the fixed files as soon as I can.

PR #515, opened 2026-05-20
Update to sky130 LibreLane `config.tcl` files and excluded cell lists

jeras:
I did not start testing yet, so I am posting this here for an early review.

**It would be probably easier to review individual commits than the full diff.**
I tried to break down the commits into a set of smaller related changes, but it is not perfect.
Each commit has an extended description that might help with the review.

As discussed before I did the following changes to TCL files:
- updated the PDK `config.tcl` file to replace deprecated variable names with new ones,
- updated SCL `config.tcl` files to replace HD/HDLL/HVL/HS/MS/LS differences with a common variable, the diff between this files is now very small, this should make future updates a bit easier
- updated SCL `config.tcl`, now globbing is used in some cell lists instead of a subset of available cells (fill, clock buffers, ...)

The second set of changes are to exclude cell lists:
- removed cells that are actually not present in the given SCL (the original files were copy pasted from the HD SCL),
- removed most sequential/combinational cells with no information why they are there (supposedly connectivity issues)
  NOTE: this might necessitate additional testcases
- added comments and using globbing to make lists short


jeras:
Just a not on what else I am working on:
- the low power LP SCL, I have a working version, but I have to rebase the changes above this branch and test again
- I updated the `config.tcl` file for OSU T12, plan to expand to T15/T18, I do not know what issues await me, so I am not sure I will be able to get it to work with LibreLane
- I found another sky130 SCL [Open Source Radation-Hardened Standard Cell Library](https://github.com/stineje/sky130RHBDlib), it is still being actively developed

RTimothyEdwards:
@jeras :  There being no real need for additional standard cell libraries, I would not recommend spending too much time on the OSU standard cell sets for sky130.

RTimothyEdwards (commented):
Looks good to me so far.
Should probably get @mole99 to look at it, especially as it affects the HD library, which is the most commonly used standard cell library.  Anything that breaks synthesis with the HD library would have a huge impact.

jeras:
Cell lists do indeed affect synthesis, I did run tests with the `spm` design and it went well, I did not run other tests yet, I hope to do it today, but for running the full test suite, using CI would be great. I also plan to update the LibreLane handling of SYNTH_EXCLUDES, which is currently done by editing Liberty files, but the list can be passed as an argument to yosys/abc.

OK, I will put the OSU SCL aside, although I might just create a PR with some unfinished changes (still an improvement on the current state).

But I will have a look at the radiation hardened library, there is an opportunity for me to learn something useful there.

jeras:
I have run some tests. I had to fix a Makefile copying the renamed excluded cell lists, otherwise the changes work.

I have run LibreLane build for the `spm`, `aes` and `picorv32` designs.

The tests were run using:
- for reference PDK: the default LibreLane 3.0.3 and
- for new PDK: using LibreLane dev branch but also modifying Nix to use the 26Q2 tag of OpenLane (and corresponding OpenSTA and ABC), I also experimented with additional cells excludes (clock and delay buffers) only for the pre-routing design repair stage (buffer insertion to fix fanout/slew/capacitance).

The `aes` is failing timing in all tests including the reference 3.0.3 version and the default GitHub CI, so the results are not relevant.

The `spm` and `picorv32` tests give similar results with the LibreLane default PDK version and this one. Regardless of PDK version there are some (many) slew and max capacitance warnings, arguably the new version seems to do marginally better (the fastest corner has no warnings in the newer PDK, all other corners do regardless of PDK). The area for the new PDK seems to be a bit smaller, mostly due to being able to use `_1` drive strength cells (removed from SYNTH_EXCLUDED cells list).

Within the tests I run there were no DRC issues, but the tests do not explicitly target cells that have been removed from the `drc_exclude.cells` list.

mole99:
Hi @jeras, really nice changes, thanks!

I'm wondering if using wildcards changes the order of the cells, which could lead to unintended behavior.

For example, the order of `CTS_CLK_BUFFERS` is important if `CTS_ROOT_BUFFER` is not set: https://openroad.readthedocs.io/en/latest/main/src/cts/README.html#id1
In our case `CTS_ROOT_BUFFER` is set, so it shouldn't matter. CTS can also select the buffers itself, so we might choose to remove `CTS_CLK_BUFFERS` completely.

Similarly, the specific FILL_CELLS and DECAP_CELLS in the HD library may have been chosen to reach the min/max density limits. I think there was such an issue back at Efabless. @RTimothyEdwards do you know anything about it?
This is relevant if one plans to tape out with e.g. ChipFoundry.

I'm wondering if we should use `buf_1` for `SYNTH_BUFFER_CELL`. Not sure what difference it makes.

FYI `sky130/librelane/sky130_fd_sc_hvl/synth_excluded.cells` is missing comments.

Before the merge, we need to run the full LibreLane CI. I don't remember how to do that locally, I'll ask in FOSSi Chat.

PS: I would also be interested in the rad-hard SCL.

RTimothyEdwards:
@mole99 : 
(1) I am going to merge your pull request and then we'll need to check if this PR is still clean with respect to the base branch.
(2) There are definitely density issues with sky130 that require use of specific fill cells.  In particular, over-use of large decap cells in otherwise empty areas can exceed the diffusion density limit, while over-use of large fill cells can violate diffusion density in the other direction.
(3) I would assume that the proper use of `SYNTH_BUFFER_CELL` is to declare the minimum size buffer, and let the sizing step beef it up where needed.  But I'm not familiar enough with OpenROAD to know if that's true or not.

jeras:
Regarding `SYNTH_BUFFER_CELL`, Yosys uses it through [`insbuf`](https://yosyshq.readthedocs.io/projects/yosys/en/latest/cmd/index_passes_techmap.html#cmd-insbuf) for something like hardened macro synthesis to connect inputs to outputs where there is no logic between them. As I understand this buffers (feedthrough) are not removed by OpenROAD [`remove_buffers`](https://openroad.readthedocs.io/en/latest/main/src/rsz/README.html#remove-buffers) command. LibreLane by default instructs OpenRoad to run [`buffer_ports -inputs`](https://github.com/librelane/librelane/blob/main/librelane/scripts/openroad/repair_design.tcl#L36-L38) and [`buffer_ports -inputs`](https://github.com/librelane/librelane/blob/main/librelane/scripts/openroad/repair_design.tcl#L40-L42) to insert additional buffers at ports. I do not know under what circumstances OpenROAD would upsize this buffer. Using `buf_1` might be safe.

Regarding `CTS_ROOT_BUFFER` and `CTS_CLK_BUFFERS` it should not be a problem since `CTS_ROOT_BUFFER` is defined. I might test how CTS behaves without an explicit list of buffers. I tried to figure out whether CTS would also insert inverters, I don't know yet, but I do know they cant be specified the way [buffers can](https://openroad.readthedocs.io/en/latest/main/src/cts/README.html#id1). I did observe what the CTS stage does, but did not check if/how further stages modify the clock tree.

When it comes to `FILL_CELLS`, `DECAP_CELLS` and other cell lists, where I changed the limited list into a glob, I can (will) rebase the commit, so the changes are explicit and properly documented, this makes it easier to revert them later or just skip them now.

I will add the comments to `sky130/librelane/sky130_fd_sc_hvl/synth_excluded.cells`.

jeras:
I rebased the commits as discussed. Cosmetic and functional changes are separated.

Globbing of cell list variables `TRISTATE_CELLS`/`CTS_CLK_BUFFERS`/`FILL_CELLS`/`DECAP_CELLS` can be reverted for each variable separately if they cause regressions.

jeras:
I fixed a typo (missing underscore `_`) in HDLL `pnr_excluded.cells`.

Added some more tristate buffers (wired OR multiplexer cells) to HDLL `config.tcl`.

jeras:
I rebased the changes on the current `master` branch and added a cleanup of dead code remaining after the recent renames of excluded cells lists for gf180mcu

mole99:
Thanks for the changes! As I said, this all looks good to me in principle, but we should really run it on the entire LibreLane CI. Not sure if Donn has made any progress with the local CI script?

As for `SYNTH_BUFFER_CELL`, I think we should go for the `_1` buffer then. OpenROAD will probably (or should) upsize the buffer if necessary.

Regarding FILL/DECAP, given Tim's comment above I would say we should keep the following in at least the `_hd` library:

```
set ::env(FILL_CELLS) "sky130_fd_sc_hd__fill_2 sky130_fd_sc_hd__fill_1"
set ::env(DECAP_CELLS) "sky130_fd_sc_hd__decap_3"
```

To be on the safe side, since you probably don't want to implement a Caravel-sized chip just to check the density right now :)

jeras:
OK, I have a draft idea on how to enable running LibreLane CI with a PDK on an `open_pdks` `dev` branch, so changes such as this commit could be tested using CI. This could speed up the review process for changes with potential regressions. I will explain this idea on the Element forum.

I already rebased the PR, so that changes to `FILL_CELLS`/`DECAP_CELLS` are isolated in a separate commit, therefore easy to revert or reapply.

RTimothyEdwards (approved):
Honestly, it all looks fine to me except I'm only a casual LibreLane user myself, so it's really all about whether or not anything breaks LibreLane.  But Leo appears to be okay with it, so I will go ahead and merge this.

RTimothyEdwards:
Pulled and merged to opencircuitdesign.com.  The github mirror will be updated overnight.

mole99:
@RTimothyEdwards Actually, I would preferred to run the changes on the LibreLane CI first 😅
In the future, I’ll approve the PR via the GitHub interface to indicate that it's ready.
But nevermind. Now that there is a Ciel release, @jeras should be able to run the changes on the entire CI and identify any issues.

> I already rebased the PR, so that changes to FILL_CELLS/DECAP_CELLS are isolated in a separate commit, therefore easy to revert or reapply.

It's definitely good that they are in separate commit. Because if what Tim said is true, we might have to revert them ^^

From Tim: 

> (2) There are definitely density issues with sky130 that require use of specific fill cells. In particular, over-use of large decap cells in otherwise empty areas can exceed the diffusion density limit, while over-use of large fill cells can violate diffusion density in the other direction.

Afaik OpenROAD uses the largest decap cells available. So this will likely lead to density issues, since we now provide all decap cells via `__decap*`.

jeras:
Will this density issues result in DRC errors, or should I look for some specific report?

I was checking today to see if the new commits are available for `fetch`ing using `ciel` they are not yet.

I will set up today a branch with the most updated branch of LibreLane (the one with Yosys 0.65) and the last PDK hash before my changes are published,  so I can use it as reference to compare with the PDK version with my changes. I also have to go through `test/design`s to see which are worth looking into for details like logic area, and slack.

I will write a report here.

RTimothyEdwards:
@jeras :  David Lindley from Chip Foundry is the person most knowledgeable about the density issues, since he's responsible for getting the Chip Foundry submissions accepted by SkyWater.  He is available on the fossi-chat server if you want to get in touch with him.

Yes, using certain cells for fill and decap can definitely cause a density DRC failure.  I was dealing with this problem myself (and later, with David Lindley) when I was working at Efabless.  As far as I know, Chip Foundry is still using a script that can do replacement on fill cells to solve a specific density problem.  For that, we created a number of `sky130_ef_sc_hd__decap_*` and `__fill_*` cells;  there are even a series of `decap_12` cells that have 20, 40, 60, and 80 percent of the normal poly area.

All of the `ef_sc_hd__decap_*12` cells have pared back the amount of local interconnect to avoid over-density issues where `decap_12` is filling large amounts of blank space in a digital design.  `fd_sc_hd__decap_12` cells have about 80 to 90% density of local interconnect inside the cell and enough of them in one place will violate local interconnect density.

`sky130_ef_sc_hd__fill_12` is basically a `decap_12` cell with no poly at all (so it is no longer a decap, but just fill).

`sky130_fd_sc_hd__fill_*` are empty inside, and there is not enough space to place diffusion and/or poly fill shapes inside the cell, so overuse of these cells in one place will cause under-density of diffusion and/or poly area.  `sky130_ef_sc_hd__fill_*` alleviates the problem with diffusion area by filling the area with islands of diffusion shapes.  This was done early on before SkyWater was insisting on a poly density standard.  Now, I think that these fill cells are going to cause poly density issues, and it is better to use decap cells instead.

David is the best person to tell you if there is an optimal set of fill and/or decap cells that works in most cases.  Ultimately, the user likely needs to be able to run full-chip density checks themselves, and be able to run the scripts that do fill cell replacement to hit the right density numbers.

mole99:
@jeras Note that we do have a density check step for KLayout in LibreLane, which is used for ihp-sg13 and gf180mcu: https://github.com/librelane/librelane/blob/30be8a38077d9bd8ad2a75041d458810a2285845/librelane/steps/klayout.py#L1141

It would be possible to add a density check step for magic by calling the appropriate script, which you can find in the installed PDK or here: https://github.com/RTimothyEdwards/open_pdks/blob/master/sky130/custom/scripts/check_density.py
Alternatively, you can run the script manually.

There has been no need for a density check on sky130 in LibreLane until now, since Efabless and now Chipfoundry have handled the filler generation and density check.

jeras:
Hi, I managed to improvise a LibreLane CI that uses `ciel` to build a PDK from a Git repo based on a hash. I used it to check this PR and 8 design tests using sky180 SCLs fail.
https://github.com/jeras/librelane/actions/runs/26859471106

I think the best approach would be to for now create an new branch (for a new PR) only containing non functional changes like renaming configuration variables, removing MUX mapping and some cleanup.

The other changes should be handled individually and more slowly (and now can be tested using CI):
1. changes to FILL and CAPS, with further consultation with David Lindley
2. changes to CTS cell lists,
3. changes to drivers/buffers,
4. changes to exclude lists, might need consultation with OpenROAD developers.

For now I only checked [one failing test](https://github.com/jeras/librelane/actions/runs/26859471106/job/79211918913#step:11:4244). The failure seems to be related to the use of `*_1` cells which I removed from the SYNTH_EXCLUDE list. The issues are slew warnings and timing failures, by removing `*_1` cells only stronger (larger) cells were used with existing PDK releases, this was enough to get a passing timing. I will have to investigate why LibreLane/OpenRoad fails to resize the cells to meet timing and slew requirements. If I had to guess (with low confidence), it would be something related to either pre-routing interconnect RC estimates being too low or post-routing parasitics being too high.

There are other issues with the current OpenROAD flow, which is inserting clock and delay buffers into data paths while repairing slew issues. I lack confidence in my understanding of the issue, so I did not file an OpenROAD issue yet.

I will first clean up my improvised CI so it can be reviewed. An additional issue is that I was unable to `ciel build` the `open_pdk` with the latest LibreLane `dev` branch (latest `nix-eda`), I will file an issue, resolving it might delay my progress on this issue.

PR #516, opened 2026-05-22
gf180mcu: write blackbox views first

mole99:
Write the blackbox views first, or else they override the Verilog models.


RTimothyEdwards (approved):
Right.  I guess I should have looked closer at the Makefile, but things like that are very hard to see without doing a full install.

RTimothyEdwards:
Pulled and merged to opencircuitdesign.com.

PR #519, opened 2026-05-28
sky130: added latches with reset

jeras:
1. Reordered ports for existing latches (`$_DLATCH_[PN]_`) to match the order in the cell Verilog definition.
2. Added latches with active low reset (`\$_DLATCH_[PN]N0_`).
3. Added cells for latches with complementary outputs, but they are commented out, since Yosys does not support them. This is done to make it explicit that additional lathes exist and that there are specific reasons for not mapping them.

~While testing this code, I noticed Yosys fails to infer latches with reset properly,
so I filled an issue~ https://github.com/YosysHQ/yosys/issues/5910
UPDATE: Yosys support for latches with preset/reset was merged.
This pull request should remain a draft, till the Yosys issue is fixed
and maybe further till the fixed version of Yosys becomes part of `nix-eda` and LibreLane.

### DIFF
--- a/sky130/librelane/sky130_fd_sc_hd/latch_map.v
+++ b/sky130/librelane/sky130_fd_sc_hd/latch_map.v
@@ -1,21 +1,74 @@
+// Delay latch, non-inverted enable, single output
 module \$_DLATCH_P_ (input E, input D, output Q);
   sky130_fd_sc_hd__dlxtp_1 _TECHMAP_DLATCH_P (
-    //# {{data|Data Signals}}
-    .D(D),
-    .Q(Q),
-
-    //# {{clocks|Clocking}}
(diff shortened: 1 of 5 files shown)


jeras:
The development version of Yosys (`v0.66+`) adds support in `dfflegalize` for more latch types (latches with preset/reset). While researching the code, I noticed it would not be too hard to implement automatic LATCH mapping, similar to how this is already done for FFs using `dfflibmap`, so I created a [PR](https://github.com/YosysHQ/yosys/pull/5976) for this feature.

The Yosys support for mapping LATCHes would make the `latch_map.v` files in `open_pdks` obsolete. So the question is, whether it is worth to maintain it, or it should be removed if/when Yosys supports it out of the box.

PR #520, opened 2026-05-30
gf180mcu: Fix PCell transistor length for 5V/6V devices

smunaut:
Obviously some cut&paste / typo.

5V NMOS are 0.60 ~ 0.695
6V NMOS are 0.70 ~ ...
5V PMOS are 0.50 ~ 0.545
6V PMOS are 0.55 ~ ...


RTimothyEdwards (approved):
I'm pretty sure that wasn't a "cut and paste" error, but one of those errors that exists between the keyboard and the chair.  Thanks for the fix.

RTimothyEdwards:
This was merged yesterday but I forgot to mention it here and close the PR, which I will do now.

PR #522, opened 2026-06-06
Update to sky130 LP port

jeras:
1. Created `config.tcl` based on the latest version for other SCLs.
2. Copied `tracks.info` from HS/MS/LS SCLs.
3. Copied (and search/replace HD with LP) Yosys technology mapping files.
4. Added clean cell exclusion lists, not just copies from HD SCL.

WARNING: the current development branch of LibreLane expects a typical corner characterization Liberty file, which is not available for LP. So it is not possible to use it out of the box with the `dev` branch. This is because the development LibreLane by default compiles the IO library which does have the typical corner Liberty files.

I still have to figure out how to test this. I did test an earlier version on LibreLane 3.0.3, and it worked, so even in this state it should be better than the current release.


RTimothyEdwards:
Have you (or anybody) tried running synthesis in LibreLane using the LP library?

jeras:
I tried (successfully) with an earlier version but not the current one, right now the main issue is that LibreLane development branch loads a pad library by default and therefore sets the list expected corners. The LP library does not have a typical (TT, 20C, 1.8V) corner characterization, so it is incompatible with the pad library. I was thinking about using some other characterization instead with the warning. I will try to run a test soon.

PR #523, opened 2026-06-07
gf180mcu: update gf180mcu_ocd_io hash

mole99:
Pull in https://github.com/RTimothyEdwards/gf180mcu_ocd_io/pull/9


RTimothyEdwards (approved):
Looks good to me, although I will just do "make update" and "make reference" on my end because it's easier.

RTimothyEdwards:
Merged the equivalent change to open_pdks on opencircuitdesign.com.

PR #526, opened 2026-06-14
magic/gf180mcu: Add {n,p}varactor to all{n,p}difflv

smunaut:


RTimothyEdwards (approved):
Thanks for the fix!

RTimothyEdwards:
Pulled and merged to opencircuitdesign.com

PR #545, opened 2026-09-17
Fix the liberty file compiler python script

### LINKED ISSUE #544: Liberty library file compilation script needs fixes
I had created my own standard cell library and characterized it using "lctime", which likes to produce single .lib files for each cell;  that's actually a reasonable output to work with, and it's exactly what the liberty file compiler python script is supposed to do, but the python script has never been used to try to compile a complete library out of individual .lib files, and it doesn't work.  The primary issue is that it removes the header from each cell but fails to remove the closing brace.  If that result is passed to yosys with "read_liberty", yosys will read the first cell, stop, and not report any error.

I will generate a pull request that fixes the problem.  It should not affect any existing use cases.

RTimothyEdwards:
This commit corrects an error in the python script that compiles liberty library files from individual per-cell liberty files, which was stripping the header from individual cells but failing to strip the closing brace at the end.  There were a handful of other minor syntax issues related to as-yet-unused command-line options which have been fixed as well.  Since there are no existing use cases which compile a liberty file from constituent cells, this change should have no effect on existing PDKs.  It is particularly useful in conjunction with characterization output produced by the tool "lctime".

Disclaimer:  This PR contains a combination of fixes by me and also by Claude Opus.  It has been tested on my work-in-progress radiation-tolerant standard cell library for GF180MCU that was characterized using "lctime".

Future work:  It could use one major additional improvement.  Currently it uses the header from the first file that it reads.  It should parse all headers and collect all matrix templates and output all templates into the single library header.  This is easy enough to work around by providing a separate header file, but that should not be necessary.


mole99 (approved):
The changes look good to me. I haven't tried them out myself, but since you're using them with lctime, that should be fine :)

Thank you for letting me know about the AI use.

RTimothyEdwards:
@mole99 :  I pushed the update before running a full PDK build and checking all the resulting liberty files, so I caught an issue and corrected it.  It's worth running a librelane synthesis check on the usual sky130_fd_sc_hd and gf180mcu7t5v0 to make sure they aren't broken.  I have checked both by "diff" and they are unchanged, so they should be okay.

RTimothyEdwards:
@mole99 :  What sort of CI is running on this?  Just a successful build in "ciel"?  The first commit I made would have broken the .lib files for both gf180mcu and sky130.  They appear to be correct now (by "diff" comparison of the .lib files) but a simple synthesis run as a check would be helpful.

mole99:
@RTimothyEdwards Whoops! I didn't think these changes would affect the other liberty files.
I built and tested sky130A/gf180mcuD successfully on LibreLane.

As for the CI, yes we're only building the PDK using Ciel. I could add another step that runs LibreLane on the PDKs, but it I'll have to find the time for that :)

RTimothyEdwards:
@mole99 :  I don't generally make changes to the main scripts for open-pdks any more, so it's probably okay just to do a quick run and check before merging.  With that done, I'll go ahead and merge the change.

